PR #20, opened 2018-11-12
filter out null geo dim

lucafuji:
Right now for geo join, we keep null geo dim values which is redundant and confusing. We need to remove those rows so that result won't contain null dim value

### DIFF
--- a/query/aql_compiler.go
+++ b/query/aql_compiler.go
@@ -280,6 +280,7 @@ func (qc *AQLQueryContext) matchGeoJoin(joinTableID int, mainTableSchema *memsto
 		pointTableID:  point.TableID,
 		pointColumnID: point.ColumnID,
 		dimIndex:      -1,
+		inOrOut:       true,
 	}
 
 	// Set column usage for geo points.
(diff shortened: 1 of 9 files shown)


query/aql_compiler.go
```diff
@@ -280,6 +280,7 @@ func (qc *AQLQueryContext) matchGeoJoin(joinTableID int, mainTableSchema *memsto
 		pointTableID:  point.TableID,
 		pointColumnID: point.ColumnID,
 		dimIndex:      -1,
+		inOrOut:       true,
```
jshencode: if we separate intersect and dimension write, we might need to add compiler check and fail geo join queries with inOrOut as false

lucafuji: We already fail it. 
see aql_compiler:677

if isCallRef && callRef.Name == geographyIntersectsCallName {
				qc.Error = utils.StackError(nil, "Not %s condition is not allowed", geographyIntersectsCallName)
				break
			}

jshencode: I see.

PR #22, opened 2018-11-15
Support missing eventtime upsert

jshencode:


memstore/ingestion.go
```diff
@@ -389,6 +413,10 @@ func writeBatchRecords(columnDeletions []bool,
 			if needToWrite {
 				vectorParty.SetDataValue(recordInfo.index, *newValue, IgnoreCount)
 			}
+
+			if isFirstEventTime {
+				batch.NumRecordsWithoutEventTime--
```
lucafuji: we can use null vector's numTrue

memstore/archiving.go
```diff
@@ -138,12 +141,24 @@ func (ss liveStoreSnapshot) createArchivingPatches(
 func (s *LiveStore) getBatchIDsToPurge(cutoff uint32) []int32 {
 	var batchIDs []int32
 	s.RLock()
+
+	s.tableSchema.RLock()
+	allowMissingEventTime := s.tableSchema.Schema.Config.AllowMissingEventTime
```
lucafuji: add validation for disallowing change AllowMissingEventTime from true to false

lucafuji:
LGTM except for the two issues I mentioned.

memstore/archiving.go
```diff
@@ -138,12 +141,24 @@ func (ss liveStoreSnapshot) createArchivingPatches(
...
 	for batchID, batch := range s.Batches {
 		if batchID >= s.LastReadRecord.BatchID {
 			continue
 		}
 
-		if _, maxValue := batch.Columns[0].(common.LiveVectorParty).GetMinMaxValue(); maxValue < cutoff {
+		timeColumn := batch.Columns[0]
+		if timeColumn != nil {
+			if _, maxValue := timeColumn.(common.LiveVectorParty).GetMinMaxValue(); maxValue < cutoff {
+				if !allowMissingEventTime || timeColumn.GetNonDefaultValueCount() == batch.Capacity || batch.MaxArrivalTime < cutoff {
```
voyager-dw: how the maxArrivalTime is chosen and updated? the no event timestamp event maybe updated with real timestamp, while some older event still has no timestamp. maybe no big issue

PR #25, opened 2018-11-16
fetch schema job impl

shz117:
- fetch schema job impl
 - ares controller client impl
 - updateTable impl in diskstore


metastore/schema_fetch.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/uber/aresdb/utils"
+	"reflect"
+	"time"
+)
+
+// SchemaFetchJob is a job that periodically pings ares-controller and updates table schemas if applicable
+type SchemaFetchJob struct {
+	clusterName      string
+	hash             string
+	interval         int
```
jshencode: maybe rename to intervalInSeconds will be more clear

clients/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return
+	}
+	var resp *http.Response
+	resp, err = c.c.Do(req)
+	if err != nil {
+		return
+	}
+	if resp.StatusCode != http.StatusOK {
+		err = utils.StackError(fmt.Errorf("controller client error fetching schema"), fmt.Sprintf("code %d", resp.StatusCode))
```
jshencode: controller error fetching schema could be just in the error message

cmd/cmd.go
```diff
@@ -152,6 +153,17 @@ func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger b
 	batchStatsReporter := memstore.NewBatchStatsReporter(5*60, memStore, metaStore)
 	go batchStatsReporter.Run()
 
+	if cfg.ClusterName != "" {
```
jshencode: also, when we are in cluster mode, schema handler should be open only on debug port

api/schema_handler.go
```diff
@@ -128,6 +131,10 @@ func (handler *SchemaHandler) GetTable(w http.ResponseWriter, r *http.Request) {
 //    default: errorResponse
 //        200: noContentResponse
 func (handler *SchemaHandler) AddTable(w http.ResponseWriter, r *http.Request) {
+	if handler.isClusterMode {
```
jshencode: we can modify Register to just register the read/Get calls

shz117: i'm afraid that would confuse user with 404 error. 405 (not allowed) may make more sense?

cmd/cmd.go
shz117: does it make sense to still open read methods?

shz117: I was considering styx/subscriber will still need to read schema from instance during migration

clients/controller.go
shz117: what do you mean?

jshencode: "controller client error" could be in the message

api/schema_handler.go
jshencode: hmm... i think we can use router.Methods() to control the methods

PR #28, opened 2018-11-27
move initial schema fetch before memstore init

shz117:


cmd/cmd.go
```diff
@@ -73,6 +73,22 @@ func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger b
...
+			logger.Fatal("Missing cluster name")
+		}
+		controllerClientCfg := cfg.Clients.Controller
+		if controllerClientCfg == nil {
+			logger.Fatal("Missing controller client config", err)
+		}
+		controllerClient := clients.NewControllerHTTPClient(controllerClientCfg.Host, controllerClientCfg.Port, controllerClientCfg.Headers)
+		schemaFetchJob := metastore.NewSchemaFetchJob(5*60, metaStore, metastore.NewTableSchameValidator(), controllerClient, cfg.Cluster.ClusterName, "")
+		// immediate initial fetch
+		schemaFetchJob.FetchSchema()
```
jshencode: also, we need to take a close look at the schema mutations to make sure it does not push to channels that have not been setup yet.

shz117: I think current pushSchemaChange already handles this case
```
func (dm *diskMetaStore) pushSchemaChange(table *common.Table) {
	if dm.tableSchemaWatcher != nil {
		dm.tableSchemaWatcher <- table
		<-dm.tableSchemaDone
	}
}
```

PR #29, opened 2018-11-30
add timeout for controller client

shz117:


common/config.go
```diff
@@ -49,6 +49,8 @@ type HTTPConfig struct {
 type ControllerConfig struct {
 	Host    string      `yaml:"host"`
 	Port    int         `yaml:"port"`
+	// Timeout for all requests in seconds. if < 0, will use default value 5
```
lucafuji: need to be consistent with code. Code is <= 0

cmd/cmd.go
```diff
@@ -85,7 +85,7 @@ func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger b
 		if cfg.Cluster.InstanceName != "" {
 			controllerClientCfg.Headers.Add(clients.InstanceNameHeaderKey, cfg.Cluster.InstanceName)
 		}
-		controllerClient := clients.NewControllerHTTPClient(controllerClientCfg.Host, controllerClientCfg.Port, controllerClientCfg.Headers)
+		controllerClient := clients.NewControllerHTTPClient(*controllerClientCfg)
```
lucafuji: actually cluster mode sounds weird for me. For now it's just remote schema not truly a cluster

shz117: soft agreed.. membership maintained by heart beat is the next thing to follow (in my current plan), that might make the word "cluster" sound better, unless the plan is overriden by something else.

PR #35, opened 2018-12-04
cache cuda deps

shz117:
- enable travis cache
 - trim CUDA deps, skip CUDA deps installation when applicable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,10 @@ env:
     - LD_LIBRARY_PATH=${CUDA_HOME}/lib64:${TRAVIS_HOME}/gopath/src/github.com/uber/aresdb/lib
     - NVCCFLAGS="-ldir ${CUDA_HOME}/nvvm/libdevice -noprof -I ${CUDA_HOME}/include/ -L ${CUDA_HOME}/lib64"
     - GTEST_ROOT=./build/googletest
+cache: 
+  directories:
+    - lib
+    - gtest
 
 cache:
   directories:
@@ -22,15 +26,22 @@ cache:
 branches:
   only:
     - master
-
 install:
   - .travis/install_cuda.sh
   - .travis/install_golang_tools.sh
   - .travis/install_clang_tools.sh
-
 script:
   - make lint
   - make travis
-
+after_success:
+  # exclude query/expr from coverage
+  - rm query/expr/expr.coverprofile
+  # generate coverage output file
+  - gover . coverage.out
+  # send to coverall
+  - goveralls -coverprofile=coverage.out -service=travis-ci 
+  # send to codecov.io
+  - mv coverage.out coverage.txt
+  - bash <(curl -s https://codecov.io/bash)
 after_script:
   - .travis/cleanup.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,16 @@ env:
     - CUDA_VERSION="10.0"
     - CUDA_PKG_VERSION="130-1"
     - CUDA_PKG=cuda-repo-ubuntu1404_${CUDA_VERSION}.${CUDA_PKG_VERSION}_amd64.deb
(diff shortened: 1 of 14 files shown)


.travis/install_cuda.sh
```diff
@@ -1,9 +1,15 @@
...
-wget http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/${CUDA_PKG}
-sudo apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/7fa2af80.pub
-sudo dpkg -i ${CUDA_PKG}
-rm ${CUDA_PKG}
-sudo apt-get -y update
-sudo apt-get install -y --no-install-recommends  cuda-core-${CUDA_VERSION}  cuda-cudart-dev-${CUDA_VERSION}  cuda-cublas-dev-${CUDA_VERSION} cuda-curand-dev-${CUDA_VERSION}
-sudo ln -sf /usr/local/cuda-${CUDA_VERSION} /usr/local/cuda
\ No newline at end of file
+
+if [ "$(ls -A $CUDA_LIB_LOCAL)" ]; then
```
lucafuji: can we simply check the existence of the directory?

.travis.yml
```diff
@@ -7,11 +7,17 @@ env:
     - CUDA_VERSION="10.0"
     - CUDA_PKG_VERSION="130-1"
     - CUDA_PKG=cuda-repo-ubuntu1404_${CUDA_VERSION}.${CUDA_PKG_VERSION}_amd64.deb
-    - CUDA_HOME=/usr/local/cuda
-    - PATH=${PATH}:${CUDA_HOME}/bin:${GOPATH}/bin
-    - LD_LIBRARY_PATH=${CUDA_HOME}/lib:${TRAVIS_HOME}/gopath/src/github.com/uber/aresdb/lib
+    - CUDA_LIB_LOCAL=${HOME}/cuda_lib
```
lucafuji: maybe better to have version in the folder name so that later when we change cuda version, we don't need to manually delete the cache.

.travis/install_cuda.sh
```diff
@@ -1,9 +1,15 @@
...
+    echo "Fetched cuda deps from cache"
+else
+    wget http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/${CUDA_PKG}
+    sudo apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/cuda/repos/ubuntu1404/x86_64/7fa2af80.pub
+    sudo dpkg -i ${CUDA_PKG}
+    rm ${CUDA_PKG}
+    sudo apt-get -y update
+    sudo apt-get install -y --no-install-recommends  cuda-core-${CUDA_VERSION}  cuda-cudart-dev-${CUDA_VERSION}  cuda-cublas-dev-${CUDA_VERSION} cuda-curand-dev-${CUDA_VERSION}
+    mkdir -p ${CUDA_LIB_LOCAL}
+    sudo cp -rL /usr/local/cuda-${CUDA_VERSION}/bin /usr/local/cuda-${CUDA_VERSION}/lib64 /usr/local/cuda-${CUDA_VERSION}/include /usr/local/cuda-${CUDA_VERSION}/nvvm ${CUDA_LIB_LOCAL}
```
lucafuji: Seems we are still copying all the libs and bins?

.travis/install_cuda.sh
shz117: we cant, since travis cache will create the dir once configured

.travis/install_cuda.sh
shz117: yes. but it's a subset of what's installed, if not cached. any other suggestion?

.travis.yml
shz117: good point, will update

shz117:
cache cuda deps

PR #38, opened 2018-12-04
add coverall in travis

voyager-dw:


.travis/run_unittest.sh
```diff
@@ -7,11 +7,4 @@ make test-cuda -j
...
-ginkgo -r
-
-echo "mode: atomic" > coverage.out
-for file in $(find . -name "*.coverprofile" ! \( -name "coverage.out"  -o -name "expr.coverprofile" \) ); do \
-    cat $file | grep -v "mode: atomic" | awk 's=index($0,"ares")+length("ares") { print "." substr($0, s)}' >> coverage.out ; \
-    #rm $file ; \
-done
-gocov convert coverage.out | gocov-xml > coverage.xml
\ No newline at end of file
+ginkgo -r -cover
```
lucafuji: we need to ignore query.expr as it's coverage is low and belongs to influx db.

voyager-dw: remove expr.coverprofile

PR #40, opened 2018-12-05
change coverage to codecov

voyager-dw:


.travis/run_unittest.sh
```diff
@@ -33,7 +33,10 @@ make test-cuda -j
 make ares -j
 
 # run test
-ginkgo -r -cover
+cd query/expr
+ginkgo -r
+cd ../..
+ginkgo -r -cover -skipPackage expr
```
lucafuji: query. expr? Seems the coverage is still 75%

PR #41, opened 2018-12-12
add distributed package

shz117:
- add distributed pkg
 - create membership manager
 - manage membership with zk ephemeral znode
 - refactor schema_fetch job into  membership manager


Makefile
```diff
@@ -242,3 +242,9 @@ $(GTEST_OUT_DIR)/%_unittest.o: $(CXX_SRC_DIR)/query/%_unittest.cu
 
 $(GTEST_OUT_DIR)/all_unittest : $(patsubst %,$(GTEST_OUT_DIR)/%.o,$(CUDA_TESTS)) $(GTEST_OUT_DIR)/gtest_main.a
 	$(NVCC) $(CPPFLAGS) $(CXXFLAGS) $(NVCCFLAGS) $^ -o $@  -lpthread -lalgorithm
+
```
voyager-dw: there is no importable zookeep client lib?

cmd/cmd.go
```diff
@@ -82,14 +82,12 @@ func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger b
 		if controllerClientCfg == nil {
```
voyager-dw: the controller is not open source yet, right?

distributed/membership.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	_, err = mm.zkc.Create(getZNodePath(clusterName, instanceName), instanceBytes, zk.FlagEphemeral, zk.WorldACL(zk.PermAll))
+	if err != nil {
+		return
+	}
+
+	// start jobs
+	controllerClient := clients.NewControllerHTTPClient(mm.cfg.Clients.Controller.Host, mm.cfg.Clients.Controller.Port, mm.cfg.Clients.Controller.Headers)
+	// TODO: (shz) rewrite schema fetch job to use zk watches
```
voyager-dw: maybe we put together for all zk related changes, otherwise will be in middle stage as this is open source project.

Makefile
shz117: we are using `github.com/samuel/go-zookeeper`. do you mean a mock?

cmd/cmd.go
shz117: not yet.
there's a way to enable/disable cluster mode, so the dependency on controller is optional.

distributed/membership.go
shz117: ok. I will add that change to this PR.

cmd/cmd.go
shz117: removed controller client, connect to zk directly instead

PR #46, opened 2019-01-11
Update Dockerfile

shz117:


Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN apt-get update
 RUN apt-get install -y golang-1.9-go git
 
 # clone aresdb repo and set up GOPATH
-RUN mkdir -p
+RUN mkdir -p aresdb
```
lucafuji: should be mkdir -p $UBER_GITHUB_DIR
Possibly I miss this after a refactor

PR #47, opened 2019-01-11
copy lock when over when creating new batch

shz117:
needs integration test

### DIFF
--- a/memstore/archive_store.go
+++ b/memstore/archive_store.go
@@ -144,7 +144,7 @@ func (v *ArchiveStoreVersion) RequestBatch(batchID int32) *ArchiveBatch {
 		Size:    size,
 		BatchID: batchID,
 		Shard:   v.shard,
-		Batch: Batch{RWMutex: &sync.RWMutex{}},
+		Batch:   Batch{RWMutex: &sync.RWMutex{}},
 	}
 	v.Batches[batchID] = batch
 	return batch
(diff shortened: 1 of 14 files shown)


memstore/archive_store.go
```diff
@@ -374,6 +374,7 @@ func (b *ArchiveBatch) BuildIndex(sortColumns []int, primaryKeyColumns []int, pk
 func (b *ArchiveBatch) Clone() *ArchiveBatch {
 	newBatch := &ArchiveBatch{
 		Batch: Batch{
+			RWMutex: b.Batch.RWMutex,
```
lucafuji: need to copy by reference. therefore we may need to change the lock to a pointer inside batch

memstore/archive_store.go
```diff
@@ -218,7 +218,7 @@ func (b *ArchiveBatch) RequestVectorParty(columnID int) common.ArchiveVectorPart
 	// columnID should always be smaller than len(ValueTypeByColumn).
 	dataType := b.Shard.Schema.ValueTypeByColumn[columnID]
 	defaultValue := b.Shard.Schema.DefaultValues[columnID]
-	vp = newArchiveVectorParty(b.Size, dataType, *defaultValue, &b.RWMutex)
+	vp = newArchiveVectorParty(b.Size, dataType, *defaultValue, b.RWMutex)
```
jshencode: should be more places that need similar change

PR #50, opened 2019-01-15
Enable JSONResponseWriter to report hll value

jshencode:


api/query_handler.go
```diff
@@ -257,7 +257,16 @@ func (w *JSONQueryResponseWriter) ReportQueryContext(qc *query.AQLQueryContext)
...
 // ReportResult writes the query result to the response.
 func (w *JSONQueryResponseWriter) ReportResult(queryIndex int, qc *query.AQLQueryContext) {
-	qc.Results = qc.Postprocess()
+	if qc.OOPK.IsHLL() {
+		result, err := queryCom.NewTimeSeriesHLLResult(qc.HLLQueryResult, queryCom.OldHLLDataHeader)
+		if err != nil {
+			// should never be here except bug
+			w.ReportError(queryIndex, qc.Query.Table, utils.StackError(err, "failed to read hll result"), http.StatusInternalServerError)
+		}
+		qc.Results = queryCom.ComputeHLLResult(result)
```
lucafuji: can we wrap those logic inside Postprocess?

query/common/hll.go
```diff
@@ -485,6 +485,32 @@ func parseTimeseriesHLLResult(buffer []byte) (AQLTimeSeriesResult, error) {
...
+		return r
+	case map[string]interface{}:
+		for k, v := range r {
+			r[k] = computeHLLResultRecursive(v)
+		}
+		return r
+	case float64:
+		return r
+	case HLL:
+		return r.Compute()
```
lucafuji: panic or raise an error when case not match?

PR #51, opened 2019-01-18
Viper and Cobra integration

jshencode:


cmd/cmd.go
```diff
@@ -55,19 +142,11 @@ func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger b
...
-	schedulerOff := utils.BoolFlag("scheduler_off", "so", utils.GetConfig().SchedulerOff, "Start server with scheduler off, no archiving and backfill")
-
-	flag.Parse()
-
 	scope.Counter("restart").Inc(1)
 	serverRestartTimer := scope.Timer("restart").Start()
 
 	// Create MetaStore.
-	metaStorePath := filepath.Join(*rootPath, "metastore")
+	metaStorePath := filepath.Join(utils.GetConfig().RootPath, "metastore")
```
voyager-dw: config is already passed in

utils/di.go
```diff
@@ -30,10 +31,21 @@ var (
...
 // init loads default implementations of common components for unit tests' purpose.
 func init() {
+	ResetDefaults()
+}
+
+// ResetDefaults reset default config, logger and metrics settings
+func ResetDefaults() {
 	logger = common.NewLoggerFactory().GetDefaultLogger()
 	queryLogger = common.NewLoggerFactory().GetDefaultLogger()
 	scope := tally.NewTestScope("test", nil)
```
voyager-dw: why test scope?

utils/config.go
```diff
@@ -0,0 +1,65 @@
...
+		"instance_name": hostname,
+	})
+	v.MergeConfigMap(defaultCfg)
+
+	// merge in config file
+	if cfgFile, err := flags.GetString("config"); err == nil && cfgFile != "" {
+		v.SetConfigFile(cfgFile)
+	} else {
+		v.SetConfigName("ares")
+		v.AddConfigPath("./config")
```
voyager-dw: how about customized config path?

utils/di.go
jshencode: these are default settings used in tests mostly

in production we set using utils.Init()

cmd/cmd.go
```diff
@@ -34,10 +33,47 @@ import (
...
 	"github.com/uber/aresdb/memutils"
+	"github.com/spf13/cobra"
 )
 
-// StartService is the entry point of starting ares.
-func StartService(cfg common.AresServerConfig, logger bark.Logger, queryLogger bark.Logger, metricsCfg common.Metrics, httpWrappers ...utils.HTTPHandlerWrapper) {
+// AppParams contains application parameters for creating command
+type AppParams struct {
+	DefaultCfg   map[string]interface{}
+	ServerLogger bark.Logger
```
lucafuji: maybe we can enhance this by checking ServerLogger, QueryLogger, Metrics is nil or not and set default accordingly

cmd/cmd.go
```diff
@@ -34,10 +33,47 @@ import (
...
+	QueryLogger bark.Logger
+	Metrics common.Metrics
```
lucafuji: instead of using a big struct, probably we can use optional argument pattern here:
https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md
so user can specify what they want and leave defaults for remaining

jshencode: Done

cmd/cmd.go
jshencode: Done

cmd/cmd.go
```diff
@@ -31,13 +30,72 @@ import (
...
+func Execute(setters ...Option) {
+
+	options := &Options{}
+	for _, setter := range setters {
+		setter(options)
+	}
+
+	loggerFactory := common.NewLoggerFactory()
+	if options.ServerLogger == nil {
+		options.ServerLogger = loggerFactory.GetDefaultLogger()
```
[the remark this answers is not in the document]
jshencode: Done

utils/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"os"
+)
+
+// BindenvsToViper binds environment variables to viper
+func bindEnvironments(v *viper.Viper) {
+	v.SetEnvPrefix("ares")
+	v.BindEnv("env")
+}
+
+// AddFlagsToCommand adds flags to command
```
shz117: nit

utils/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"fmt"
+	"github.com/mitchellh/mapstructure"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"github.com/uber/aresdb/common"
+	"os"
+)
+
+// BindenvsToViper binds environment variables to viper
```
shz117: nit function name

utils/config.go
jshencode: Done

utils/config.go
jshencode: Done

PR #59, opened 2019-01-29
Bark to zap

lxning:
This diff is to migrate bark to zap and clean up log code.

### DIFF
--- a/memstore/recovery.go
+++ b/memstore/recovery.go
@@ -17,11 +17,11 @@ package memstore
 import (
 	"sync"
 
-	"github.com/uber-common/bark"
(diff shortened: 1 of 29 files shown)


memstore/recovery.go
```diff
@@ -66,7 +66,7 @@ func (shard *TableShard) ReplayRedoLogs() {
 		shard.LiveStore.WriterLock.Unlock()
 
 		if err != nil {
-			utils.GetLogger().WithError(err).Panic("Failed to apply upsert batch during recovery")
+			utils.GetLogger().With(err).Panic("Failed to apply upsert batch during recovery")
```
lucafuji: with(LogFields{"err":error}))

PR #63, opened 2019-01-30
Install glide and swagger in setup phase

rShetty:
- Use setup in Makefile for download dependencies required for building the binary.


rShetty:
Also it is better if we move to `go mod` as it is now the accepted standard for building dependencies over `glide`, let me know if it is ok to move to `go mod`, I can raise a MR on this. Thanks

PR #69, opened 2019-01-30
fix vet

shz117:


memstore/table_shard.go
```diff
@@ -62,7 +62,7 @@ func NewTableShard(schema *TableSchema, metaStore metastore.MetaStore,
 		HostMemoryManager: hostMemoryManager,
 	}
 	archiveStore := NewArchiveStore(tableShard)
-	tableShard.ArchiveStore = archiveStore
+	tableShard.ArchiveStore = &archiveStore
```
lucafuji: maybe better just have NewArchiveStore return a pointer

PR #73, opened 2019-01-31
refactor for proper defer

shz117:


shz117:
in old implementation, defer resource release within for loop will be executed outside for loop, this might starve queries (after first allocated one) in the same request.
this diff makes sure devices are released after processing each query.

PR #74, opened 2019-02-01
introduce HLLConfig

jshencode:
1. HLLConfig with two modes:
    hllOnly:  column is only used for hll aggregation
    hllEnabled: extract column will be created for hll aggregation

2. schema creation will automatically create hll column according to hll config
    default column suffix: "hll", user can override suffix in hllconfig

3. ingestion client will directly compute and insert hllValue

4. countDistinctHLL will determine whether to calculate hll value on the fly based on hllconfig
    hllOnly: directly apply hll function to column
    hllEnabled: directly apply hll function to the derived hll column
    nothing set: query engine will compute hll value on the fly and apply hll function


Makefile
```diff
@@ -7,7 +7,7 @@ ALL_GO_SRC := $(shell find . -name "*.go" | grep -v -e Godeps -e vendor -e go-bu
   -e ".*/_.*" \
   -e ".*/mocks.*")
 
-CHANGED_GO_SRC := $(shell git diff HEAD origin/master --name-only | grep -e ".*\.go")
+CHANGED_GO_SRC := $(shell git diff origin/master HEAD --name-only --diff-filter=AM | grep -e ".*\.go")
```
shz117: `--diff-filter=AMR` so we also lint renamed files

client/connector.go
```diff
@@ -198,6 +199,56 @@ func (c *connector) Insert(tableName string, columnNames []string, rows []Row, u
...
+	var out [2]uint64
+	var ok bool
+	numBytes := 4
+	switch dataType {
+	case memCom.UUID:
+		out, ok = memCom.ConvertToUUID(value)
+		if ok {
+			return out[0] ^ out[1], nil
+		}
+	case memCom.Uint32:
```
voyager-dw: better have a common convert function instead of switch here

client/connector.go
```diff
@@ -198,6 +199,56 @@ func (c *connector) Insert(tableName string, columnNames []string, rows []Row, u
...
+		value, ok = memCom.ConvertToUint32(value)
+		numBytes = 4
```
voyager-dw: bytes should come from the data type calculation DataType & 0x0000FFFF: The width of the data type in bits.

client/connector.go
```diff
@@ -198,6 +199,56 @@ func (c *connector) Insert(tableName string, columnNames []string, rows []Row, u
...
+	case memCom.Int64:
+		value, ok = memCom.ConvertToInt64(value)
+		numBytes = 8
+	default:
+		return 0, utils.StackError(nil, "Unsupported data type for hyperloglog computation")
+	}
+	if !ok {
+		return 0, utils.StackError(nil, "Invalid %v for data type %s", memCom.DataTypeName[dataType])
+	}
+	out = utils.Murmur3Sum128(unsafe.Pointer(&value), numBytes,  0)
```
voyager-dw: do we only use first 64 bit? then just return uint64?

client/connector.go
jshencode: Done

client/connector.go
jshencode: Done

client/connector.go
jshencode: Done

Makefile
jshencode: Done

client/connector.go
```diff
@@ -198,6 +199,37 @@ func (c *connector) Insert(tableName string, columnNames []string, rows []Row, u
...
+}
+
+// computeHLLValue populate hyperloglog value
+func computeHLLValue(dataType memCom.DataType, value interface{}) (uint32, error) {
+	hashed, err := hllHash(dataType, value)
+	if err != nil {
+		return 0, err
+	}
+	p := uint32(14) // max 16
+	group := uint32(hashed & ((1 << p) - 1))
```
lucafuji: why not directly do uint32(hashed & ((1 << 14) - 1))

memstore/schema.go
```diff
@@ -151,7 +151,7 @@ func (t *TableSchema) SetDefaultValue(columnID int) {
 		if !ok {
 			// Should no happen since the enum value should already be created.
 			utils.GetLogger().With(
-				"data_type", t.Schema.Columns[columnID].Type,
+				"data_type", memCom.DataTypeName[dataType],
```
lucafuji: can we cache the dataTypeName since it's used multiple times

metastore/validator.go
```diff
@@ -39,6 +40,17 @@ func (v tableSchemaValidatorImpl) Validate() (err error) {
 	return v.validateSchemaUpdate(v.newTable, v.oldTable)
 }
 
+// ValidateHLLConfig validates hll config
+func validateColumnHLLConfig(c common.Column) error {
+	if c.HLLConfig.IsHLLColumn {
+		if c.Type != common.Uint32 && c.Type != common.Int32 && c.Type != common.Int64 && c.Type != common.UUID {
```
lucafuji: is just column width >= 4 but not float and double?

client/connector.go
```diff
@@ -307,16 +340,24 @@ func (c *connector) prepareUpsertBatch(tableName string, columnNames []string, u
...
+			// compute hll value to insert
+			if column.HLLConfig.IsHLLColumn {
+				// here use original column data type to compute hll value
+				value, err = computeHLLValue(memCom.DataTypeFromString(column.Type), value)
+				if err != nil {
+					upsertBatchBuilder.RemoveRow()
+					c.logger.With("name", "prepareUpsertBatch", "error", err.Error(), "table", tableName, "columnID", columnID, "value", value).Error("Failed to set value")
+					break
+				}
+			} else {
```
lucafuji: seems else branch is not correctly set the value?

query/aql_compiler.go
```diff
@@ -1037,11 +1038,18 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
...
 			}
+
 			e.Name = hllCallName
-			e.Args[0] = &expr.UnaryExpr{
-				Op:       expr.GET_HLL_VALUE,
-				Expr:     colRef,
-				ExprType: expr.Unsigned,
+			if colRef.IsHLLColumn {
+				// 1. column is hll column itself
+				e.Args[0] = colRef
```
lucafuji: seems this branch is a noop. Is e.Args[0] already colRef?

client/connector.go
jshencode: i actually moved this function in to utils and created a const for them

memstore/schema.go
jshencode: Done

client/connector.go
jshencode: Done

query/aql_compiler.go
jshencode: Done

metastore/common/model.go
```diff
@@ -59,6 +59,16 @@ type Column struct {
 
 	// Mutable column configs.
 	Config ColumnConfig `json:"config,omitempty"`
+
+	// HLLEnabled determines whether a column is enabled for hll cardinality estimation
+	// HLLConfig is immutable
+	HLLConfig HLLConfig `json:"hllConfig,omitempty"`
```
shz117: flatten to single bool?

metastore/validator.go
shz117: nit: maybe create a const with supported types. use `IndexOfInt` for lookup.

query/aql_compiler.go
```diff
@@ -1583,11 +1589,11 @@ func (qc *AQLQueryContext) processTimeFilter() {
 				qc.Query.TimeFilter.Column)
 			return
 		}
-		timeColumnType := qc.TableScanners[0].Schema.Schema.Columns[timeColumnID].Type
-		if timeColumnType != metaCom.Uint32 {
+		timeColumnType := qc.TableScanners[0].Schema.ValueTypeByColumn[timeColumnID]
```
shz117: why don't we should delete `metastore/common/data_type.go`, and move everything to `memstore/common`, it only defines string constants. `memstore/common` should be safe to import from anywhere.

metastore/common/model.go
jshencode: reserved for future configs

query/aql_compiler.go
jshencode: let's revisit this in a different issue

PR #78, opened 2019-02-01
reorganize binary and file location

lucafuji:
1.  Move all docker files to docker directory
2. Move generated binary to bin directory
3. rename ares to aresd
4. when run server, write the log to a file in docker


docker/README.md
```diff
@@ -20,8 +20,8 @@ $ docker build -t aresdb:latest aresdb-docker
 ### Run AresDB
 ```
 nvidia-docker run -p 9374:9374/tcp -p 43202:43202/tcp -it aresdb:latest
-root@9e4c5150659c:~/go/src/github.com/aresdb# make run_server&
+root@9e4c5150659c:~/go/src/github.com/aresdb# make run_server >> ./log/aresdb.log 2>&1 &
```
shz117: I didn't test but we might lose log along with the container.
maybe we should attach volume for log as described [here](https://stackoverflow.com/questions/26677860/how-can-i-persist-my-logs-data-to-local-filesystem-in-docker)

PR #79, opened 2019-02-01
increase tests for controller integration

shz117:


metastore/schema_fetch_test.go
```diff
@@ -88,4 +89,70 @@ var _ = ginkgo.Describe("schema fetch job", func() {
 		mockSchemaValidator.On("Validate").Return(nil)
 		job.FetchSchema()
 	})
+
+	ginkgo.It("run and stop should work", func() {
+		go job.Run()
+		job.Stop()
```
lucafuji: is there any result or intermediate state you can check for the successful run of the job?

PR #80, opened 2019-02-02
fix typo

garrusd:


cmd/cmd.go
```diff
@@ -151,7 +151,7 @@ func start(cfg common.AresServerConfig, logger common.Logger, queryLogger common
 	schemaHandler := api.NewSchemaHandler(metaStore)
 
 	// create enum handler
-	enumHander := api.NewEnumHandler(memStore, metaStore)
+	enumHandler := api.NewEnumHandler(memStore, metaStore)
```
lucafuji: please also rename HttpWrappers to HTTPWrappers since it's causing lint error, after this I think you are good to go

garrusd: done

PR #86, opened 2019-02-06
skip records with arrival time older than ArchivingCutoffHighWatermark

jshencode:


memstore/ingestion.go
```diff
@@ -206,6 +206,9 @@ func (shard *TableShard) insertPrimaryKeys(primaryKeyColumns []int, eventTimeCol
 				return nil, nil, nil, utils.StackError(err, "Event time for row %d is null", row)
 			}
 			primaryKeyEventTime = upsertBatch.ArrivalTime
+			if primaryKeyEventTime < shard.LiveStore.ArchivingCutoffHighWatermark {
```
voyager-dw: it seems for normal event, we put into backfill queue, is there any other purge time that we even dont need put into backfill queue, or handled by backfill?

if this is for yesterday's issue, we need to fix in redolog recovery too

jshencode: 1.  in total we have two cases:
    event with valid **event time**, we place into backfill queue once **event time** is older than archiving cutoff
    event with invalid **event time**, we ignore the event when **arrival time** is older than archiving cutoff 
2. this code is shared for recovery and ingestion

voyager-dw: I think yesterday's case the event time is valid, but older than the cutoff time when recover, no?

shz117: I think if it hits this branch the records `!isEventTimeValid`

voyager-dw: i see, both ingestion and recovery using this function

lucafuji: add comments saying that those missing event time records will not go to backfill queue if they are late

lucafuji: they can be combined into the previous line
```
if  primaryKeyEventTime = upsertBatch.ArrivalTime; primaryKeyEventTime < shard.LiveStore.ArchivingCutoffHighWatermark {
```

jshencode: done

jshencode: done

PR #93, opened 2019-02-07
provide default table configs

### LINKED ISSUE #90: schema creation with default values and validation of configs
**Describe the bug**
currently, user create schema through api and did not fill in table configs will produce invalid table schema and cause server panic once trying to ingest data until server restart

**To Reproduce**
1. create table through api without table config
2. ingest data
3. restart server

**Expected behavior**
panic at step 2
return normal at step 3

jshencode:
fix #90

### DIFF
--- a/api/schema_handler.go
+++ b/api/schema_handler.go
@@ -128,6 +128,8 @@ func (handler *SchemaHandler) GetTable(w http.ResponseWriter, r *http.Request) {
 func (handler *SchemaHandler) AddTable(w http.ResponseWriter, r *http.Request) {
 
 	var addTableRequest AddTableRequest
+	// add default table configs first
+	addTableRequest.Body.Config = metastore.DefaultTableConfig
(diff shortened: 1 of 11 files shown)


api/schema_handler.go
```diff
@@ -128,6 +128,8 @@ func (handler *SchemaHandler) GetTable(w http.ResponseWriter, r *http.Request) {
 func (handler *SchemaHandler) AddTable(w http.ResponseWriter, r *http.Request) {
 
 	var addTableRequest AddTableRequest
+	// add default table configs first
+	addTableRequest.Body.Config = metastore.DefaultTableConfig
```
shz117: maybe add a  schema_handler unit test  for this case

jshencode: unit test added

jshencode:
@shz117 added validation to table config struct

PR #94, opened 2019-02-07
fix schema defaults

shz117:
for issue #90


api/schema_request.go
```diff
@@ -51,7 +51,7 @@ type UpdateTableConfigRequest struct {
 	// in: path
 	TableName string `path:"table" json:"table"`
 	// in: body
-	Body metaCom.TableConfig `body:""`
+	Body *metaCom.TableConfig `body:"",omitempty`
```
voyager-dw: what is the context to make this change here?

shz117:
duplicated with #93

PR #101, opened 2019-02-25
Rename clients to gateways

shz117:


cmd/cmd.go
```diff
@@ -31,7 +31,7 @@ import (
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"github.com/spf13/cobra"
-	"github.com/uber/aresdb/clients"
```
lucafuji: according to https://rakyll.org/style-packages/
golang pkg better not to be plurals

PR #102, opened 2019-03-02
Support multiple measures

shz117:
compiler changes for multiple measures and non-agg query
processor changes for support multiple measures in one query


.gitignore
```diff
@@ -17,6 +17,7 @@ bower_components
 node_modules
 CMakeCache.txt
 CMakeFiles/
```
voyager-dw: these should not be checked in?

query/aql_compiler.go
```diff
@@ -1326,47 +1326,16 @@ func (qc *AQLQueryContext) matchPrefilters() {
...
-	if len(qc.Query.Measures) != 1 {
-		qc.Error = utils.StackError(nil, "expect one measure per query, but got %d",
-			len(qc.Query.Measures))
-		return
-	}
-
 	// Categorize common filters and prefilters based on matched prefilters.
-	commonFilters := qc.Query.Measures[0].filters
+	commonFilters := []expr.Expr{}
+	for _, measure := range qc.Query.Measures {
```
voyager-dw: should be no measure level filter if multiple measures are reqested

.gitignore
shz117: yeah these are generated by cmake which we don't need. we only need `Makefile` for now

query/time_series_aggregate.go
```diff
@@ -649,27 +650,31 @@ func (bc *oopkBatchContext) hll(numDims common.DimCountsPerDimWidth, isLastBatch
...
-func (bc *oopkBatchContext) sortByKey(numDims common.DimCountsPerDimWidth, valueWidth int, stream unsafe.Pointer, device int) {
+func (bc *oopkBatchContext) sortByKey(numDims common.DimCountsPerDimWidth, valueWidths []int, stream unsafe.Pointer, device int) {
 	keys := makeDimensionColumnVector(bc.dimensionVectorD[0].getPointer(), bc.hashVectorD[0].getPointer(),
 		bc.dimIndexVectorD[0].getPointer(), numDims, bc.resultCapacity)
-	values := (*C.uint8_t)(bc.measureVectorD[0].getPointer())
-	doCGoCall(func() C.CGoCallResHandle {
-		// sort the previous result with current batch together
-		return C.Sort(keys, values, (C.int)(valueWidth), (C.int)(bc.resultSize+bc.size), stream, C.int(device))
-	})
+	for measureIndex, valueWidth := range valueWidths {
```
voyager-dw: sort should always one time no matter how many measures

query/aql_compiler.go
```diff
@@ -1712,123 +1769,80 @@ func (qc *AQLQueryContext) processMeasureAndDimensions() {
 			aggregate.Name, len(aggregate.Args))
 		return
 	}
-	qc.OOPK.Measure = aggregate.Args[0]
+
+	measure := aggregate.Args[0]
+	var aggregateType C.enum_AggregateFunction
```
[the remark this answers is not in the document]
shz117: we will not need it since the processor will use different executor implementation for non-agg queries, right?

query/aql_compiler_test.go
```diff
@@ -1376,6 +1388,7 @@ var _ = ginkgo.Describe("AQL compiler", func() {
 			Table: "trips",
 			Measures: []Measure{
 				{Expr: "sum(fare)"},
+				{Expr: "count(*)"},
```
lucafuji: reject this query

query/aql_processor.go
```diff
@@ -1244,7 +1259,9 @@ func (qc *AQLQueryContext) estimateMemUsageForBatch(firstColumnSize int, columnM
 	memUsage += firstColumnSize * qc.OOPK.DimRowBytes * 2
 
 	// 9. Measure vector memory usage (input + output)
-	memUsage += firstColumnSize * qc.OOPK.MeasureBytes * 2
+	for _, measureBytes := range qc.OOPK.MeasureBytes {
+		memUsage += firstColumnSize * measureBytes * 2
```
lucafuji: let's move those changes to another diff and check in the non agg change first

shz117:
abandon this diff, will send in smaller batches

PR #105, opened 2019-03-07
using project(aresdb language CUDA) instead of using FindCUDA

lucafuji:
FindCUDA is a legacy way of compiling CUDA code, instead of we should using CUDA as a first class language in cmake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,6 @@ cmake_minimum_required(VERSION 3.12)
 # This has to be specified before enable_language according to
 # https://gitlab.kitware.com/cmake/cmake/issues/17559
 set( CMAKE_CUDA_FLAGS "" CACHE STRING "" )
-message(CMAKE_CUDA_FLAGS is ${CMAKE_CUDA_FLAGS})
 if ( CMAKE_CUDA_FLAGS )
 	list(REMOVE_ITEM CMAKE_CUDA_FLAGS "-cudart static")
 endif()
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,23 @@
 cmake_minimum_required(VERSION 3.12)
-find_package(CUDA REQUIRED)
+# This has to be specified before enable_language according to
+# https://gitlab.kitware.com/cmake/cmake/issues/17559
```
voyager-dw: what is the context here? We use NVCC_FLAGS from CUDA package before, and we now remove that dependency?

lucafuji: the context is the title, for reference, refer to
https://cliutils.gitlab.io/modern-cmake/chapters/packages/CUDA.html

query/iterator_unittest.cu
```diff
@@ -504,8 +503,7 @@ TEST(MeasureIteratorTest, CheckMax) {
 // cppcheck-suppress *
 TEST(RecordIDJoinIteratorTest, CheckIterator) {
   RecordID recordIDsH[5];
-  thrust::host_vector<ForeignTableIterator<int32_t>> vpItersHost(5);
-  ForeignTableIterator<int32_t> *vpIters = vpItersHost.data();
+  ForeignTableIterator<int32_t> vpItersH[5];
```
shz117: are these test changes related to the cmake change?

lucafuji: Some of the test changes are, like the bugs we have seen last time where nvcc hangs when compilation. 
Some of the test changes are just refactoring the code

PR #108, opened 2019-03-12
compiler and result change for non-agg query

shz117:


query/aql_compiler.go
```diff
@@ -1775,7 +1784,9 @@ func (qc *AQLQueryContext) processMeasureAndDimensions() {
 			"unsupported aggregate function: %s", aggregate.Name)
 		return
 	}
+}
 
+func (qc *AQLQueryContext) processDimensions() {
```
voyager-dw: how about no dimension or dimension is "*"?

query/common/non_aggregate_query_result.go
```diff
@@ -0,0 +1,50 @@
...
+const MatrixDataKey = "matrixData"
+const HeadersKey = "headers"
+
+// AQLNonAggregatedQueryResult represents non-aggregated query result
+// - there will be a "headers" key, value will be a list of column names
+// - there will be a "matrixData" key, value will be a 2d arary of values (row formated)
+type AQLNonAggregatedQueryResult map[string]interface{}
+
+// Append appends single value to specific row, result rows must be built in order
+func (r AQLNonAggregatedQueryResult) Append(row int, value interface{}) error {
```
voyager-dw: where to use this new type to return to caller? how is the original aqltimeseriesresult?

shz117: supposedly in PostProcess and handler. for non-agg query use AQLNonAggregatedQueryResult other wise use time-series-result. we might need to use `map[string]interface{}` in those places and cast these 2 types to map.
another option is to squash everything in the same general result object. wdyt?

query/aql_compiler.go
```diff
@@ -1786,12 +1786,28 @@ func (qc *AQLQueryContext) processMeasure() {
...
+		columns = append(columns, rewrittenVarRef)
+	}
+	return
+}
+
 func (qc *AQLQueryContext) processDimensions() {
 	// Copy dimension ASTs.
-	qc.OOPK.Dimensions = make([]expr.Expr, len(qc.Query.Dimensions))
-	for i, dim := range qc.Query.Dimensions {
+	qc.OOPK.Dimensions = []expr.Expr{}
```
voyager-dw: how about after we detect it is a non-aggregation query and without dimensions set, then we just add columns into dimension list and then let the compiler automatically handle all other stuffs.

query/common/time_series_result.go
```diff
@@ -73,3 +90,36 @@ func (r AQLTimeSeriesResult) SetHLL(dimValues []*string, hll HLL) {
 		}
 	}
 }
+
+// =====  Time series result methods end =====
+
+// =====  Non aggregate query result methods start =====
+
+// Append appends single value to specific row, result rows must be built in order
+func (r AQLQueryResult) Append(row int, value interface{}) error {
```
voyager-dw: likely we only need the interface to add row by row

shz117: changed

query/aql_compiler.go
shz117: I changed the implementation to expand wildcard at qc.Query.Dimension level. so processDimensions will remain unchanged.

query/common/non_aggregate_query_result.go
shz117: merged the methods to 1 struct and renamed to `AQLQueryResult`

PR #109, opened 2019-03-12
gtest and rmm as thirdparty submodule

lucafuji:


shz117:
if you already use ExternalProject why not use recommended (last) method mentioned in the gtest doc https://github.com/google/googletest/blob/master/googletest/README.md

lucafuji:
> if you already use ExternalProject why not use recommended (last) method mentioned in the gtest doc https://github.com/google/googletest/blob/master/googletest/README.md

We probably will install all 3rd parties as git submodule (like how mapd did), in such case we probably will directly use a source_dir instead of git in ExternalProject_Add.

Other than this difference, I don't see any other fundamental difference between their recommendation and ours.

PR #113, opened 2019-03-22
rmm integration

lucafuji:
1. integrate with rmm
2. unified all device memory access in c++ code. 
3. refactor deviceAllocator


memutils/memory.h
```diff
@@ -16,42 +16,72 @@
...
 #include <stddef.h>
+#include <stdint.h>
 #include "../cgoutils/utils.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+enum {
+  DEVICE_MEMORY_IMPLEMENTATION_FLAG = 1,
```
voyager-dw: HOST_MEMORY=0,
DEVICE_MEMORY=1,
DEVICE_POOLED_MEMORY=2?

memutils/memory/cuda_malloc.cu
```diff
@@ -141,17 +150,57 @@ extern "C" CGoCallResHandle GetDeviceGlobalMemoryInMB(int device) {
...
-extern "C" CGoCallResHandle CudaProfilerStop() {
+CGoCallResHandle CudaProfilerStop() {
   CGoCallResHandle resHandle = {NULL, NULL};
   cudaDeviceSynchronize();
   cudaProfilerStop();
   resHandle.pStrErr = checkCUDAError("cudaProfilerStop");
   return resHandle;
 }
+
+CGoCallResHandle GetDeviceMemoryInfo(size_t *freeSize, size_t *totalSize,
```
voyager-dw: we can return device level memory info as well for non-rmm managed

memutils/memory/rmm_alloc.cu
```diff
@@ -0,0 +1,258 @@
...
+#include <cstring>
+
+#include "../memory.h"
+
+const int MAX_ERROR_LEN = 100;
+
+// checkCUDAError checks the cuda error of last runtime calls and returns the
+// pointer to the buffer of error message. This buffer needs to be released
+// by caller or upper callers.
+char *checkCUDAError(const char *message) {
```
shz117: is this same as the one defined in cuda_malloc? possible to DRY?

memutils/memory/rmm_alloc.cu
```diff
@@ -0,0 +1,258 @@
...
+      return resHandle;
+    }
+  }
+  return resHandle;
+}
+
+CGoCallResHandle DeviceAllocate(size_t bytes, int device) {
+  CGoCallResHandle resHandle = {NULL, NULL};
+  cudaSetDevice(device);
+  // For now use default stream to avoid changing the memory allocation
```
shz117: this means deviceAlloc is still sync (for now)?

shz117: if it's an interface, does it make sense to create an abstract class that has these methods virtual?

shz117: nvm if the goal is to replace existing code, dont bother

query/device_allocator.go
```diff
@@ -20,16 +20,17 @@ import (
...
 
-//// deviceAllocatorImpl virtually allocates devices to queries.
+//// memoryTrackingDeviceAllocatorImpl virtually allocates devices to queries.
 //// It maintains device config and current query usage, allocates one or two
 //// devices to a query upon request. All requests to the GPU devices must go
 //// through this allocator in order for the virtual allocation to be effective.
 //// Two devices are allocated at the same time for a single query when consumer
 //// grade GPUs without ECC memory are used (to cross check for errors manually).
-//type deviceAllocatorImpl interface {
+//type memoryTrackingDeviceAllocatorImpl interface {
```
voyager-dw: DeviceAllocator?

query/device_allocator.go
```diff
@@ -50,7 +51,7 @@ import (
 //	//   }
 //	//   // Keep the query result only and free up most memory.
 //	//   DeviceRealloc(bytes=result_only)
-//	//   // deviceAllocatorImpl will remember this query's intent for bigger memory.
+//	//   // memoryTrackingDeviceAllocatorImpl will remember this query's intent for bigger memory.
```
voyager-dw: DeviceAllocator?

query/device_allocator.go
```diff
@@ -100,20 +102,22 @@ type deviceAllocator interface {
...
 	deviceFree(dp devicePointer)
+	// getAllocatedMemory returns allocated memory for a specific device.
+	getAllocatedMemory(device int) int64
 }
 
 var da deviceAllocator
 var deviceAllocatorOnce sync.Once
 
-// getDeviceAllocator returns singleton deviceAllocatorImpl instance.
+// getDeviceAllocator returns singleton memoryTrackingDeviceAllocatorImpl instance.
```
voyager-dw: DeviceAllocator? now there are multiple implementations

query/device_allocator.go
```diff
@@ -100,20 +102,22 @@ type deviceAllocator interface {
...
 func getDeviceAllocator() deviceAllocator {
 	deviceAllocatorOnce.Do(func() {
 		da = newDeviceAllocator()
 	})
 	return da
 }
 
-// deviceAllocate is the wrapper of deviceAllocate of deviceAllocatorImpl.
+// deviceAllocate is the wrapper of deviceAllocate of memoryTrackingDeviceAllocatorImpl.
```
voyager-dw: same as above

memutils/memory/rmm_alloc.cu
lucafuji: yes they are synced per device at host, but they will not cause device synchronization if the memory is already pooled.

query/device_allocator.go
lucafuji: yep, it's due to the refactor I will rename them

memutils/memory.h
lucafuji: we will use each bit of the flag to tell some boolean values, so we will not use continue value like here.

memutils/memory/cuda_malloc.cu
lucafuji: this cuda_malloc.cu will be deprecated after rmm integration is proven. Before that, may be it's better to rely on our memory tracking on golang side.

PR #116, opened 2019-03-27
fix non-aggr-query size calculation and add type validation

voyager-dw:


query/aql_compiler.go
```diff
@@ -1799,7 +1799,8 @@ func (qc *AQLQueryContext) processMeasure() {
 func (qc *AQLQueryContext) getAllColumnsDimension() (columns []Dimension) {
 	// only main table columns wildcard match supported
 	for _, column := range qc.TableScanners[0].Schema.Schema.Columns {
-		if !column.Deleted {
+		// TODO we can add GeoPoint data type support
+		if !column.Deleted && column.Type != "GeoPoint" && column.Type != "GeoShape" {
```
shz117: we can use metastore.common.GeoPoint const here

query/aql_compiler_test.go
```diff
@@ -3392,6 +3392,91 @@ var _ = ginkgo.Describe("AQL compiler", func() {
...
+			},
+			Measures: []Measure{
+				{
+					Expr: "1",
+				},
+			},
+		}
+		qc.parseExprs()
+		Ω(qc.Error).Should(BeNil())
+		Ω(qc.Error).Should(BeNil())
```
shz117: dup

query/aql_compiler_test.go
```diff
@@ -3392,6 +3392,91 @@ var _ = ginkgo.Describe("AQL compiler", func() {
...
+		qc.matchPrefilters()
+		qc.processFilters()
+		Ω(qc.Error).Should(BeNil())
+		qc.processMeasure()
+		Ω(qc.Error).Should(BeNil())
+		qc.processDimensions()
+		Ω(qc.Error).ShouldNot(BeNil())
+
+		qc.Error = nil
+		q = &AQLQuery{
```
shz117: qc.Query = ...

PR #119, opened 2019-03-28
Add SQL grammar

shz117:


query/sql/SqlBase.g4
```diff
@@ -0,0 +1,592 @@
+/*
```
voyager-dw: need to check with Ze/Jian on how the license works here

query/sql/antlrgen/SqlBase.interp
```diff
@@ -0,0 +1,494 @@
+token literal names:
```
voyager-dw: same as above

query/sql/SqlBase.g4
shz117: got OK from OSS legal

PR #122, opened 2019-04-02
Subscriber etcd

lxning:


subscriber/common/job/controller.go
```diff
@@ -150,6 +166,51 @@ func NewController(params Params) *Controller {
...
+		SetPlacementInstance(pInstance)
+
+	err := servicesClient.SetMetadata(sid, services.NewMetadata().
+		SetHeartbeatInterval(*params.ServiceConfig.HeartbeatConfig.Interval).
+		SetLivenessInterval(*params.ServiceConfig.HeartbeatConfig.Timeout))
+	if err != nil {
+		return err
+	}
+
+	err = servicesClient.Advertise(ad)
```
voyager-dw: can you explain a little on the usage of heartbeatservice here? to etcd? and what will do if heartbeat lost?

subscriber/common/job/controller.go
```diff
@@ -129,6 +134,17 @@ func NewController(params Params) *Controller {
 
 	if params.ServiceConfig.ControllerConfig.Enable {
 		params.ServiceConfig.Logger.Info("aresDB Controller is enabled")
+
+		if *params.ServiceConfig.HeartbeatConfig.Enabled {
+			controller.etcdServices, err = createEtcdServices(params)
```
voyager-dw: connectEtcdService?

lxning: 1. params.ServiceConfig.HeartbeatConfig.Enabled is a temporary variable to differentiate zk and etcd. Eventually, it will be removed once zk is gone.
2. connectEtcdService is used to create etcd services client.

subscriber/common/job/controller.go
lxning: heartbeatservice is an etcd lease node (kind of similar zk ephemeral node). The key will be removed if hb is lost.

etcd implements retry function internally (kind of curator functionality)

subscriber/common/job/controller.go
```diff
@@ -150,6 +166,51 @@ func NewController(params Params) *Controller {
...
 }
 
+func connectEtcdServices(params Params) (services.Services, error) {
+	iopts := instrument.NewOptions().
+		SetZapLogger(params.ServiceConfig.Logger).
+		SetMetricsScope(params.ServiceConfig.Scope)
+
+	// etcd key format: prefix/${env}/namespace/service/instanceId
+	params.ServiceConfig.EtcdConfig.Service = fmt.Sprintf("%s/%s",
+		config.ActiveAresNameSpace, params.ServiceConfig.EtcdConfig.Service)
```
voyager-dw: any typo? or reassign?

lxning: reassign EtcdConfig.Service which is defined in m3db

PR #136, opened 2019-04-17
add interactive shell

shz117:


cmd/arescli/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// ishell shell
+	shell := ishell.New()
+
+	shell.Println("Welcome to AresDB Cli!")
+	shell.AddCmd(&ishell.Cmd{
+		Name: "show",
+		Help: "`show tables` will show all tables in current cluster",
+		Func: func(c *ishell.Context) {
+			hint := "did you mean `show tables`?"
```
voyager-dw: better put the function in real separate function instead of putting all functions in one huge function

PR #137, opened 2019-04-17
upsert batch level enum dict support in connector

jshencode:
1. remove enum case preparation (no need to extend enum from connector side)
2. remove fetchAllEnumCases, enum case fetch will be included in table schema fetch

### DIFF
--- a/client/connector.go
+++ b/client/connector.go
@@ -83,10 +83,12 @@ type connector struct {
 	// use columnID instead of name since column name can be reused
 	// table names can be reused as well, deleting and adding a new table
 	// will anyway requires job restart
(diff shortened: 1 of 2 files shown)


client/connector.go
```diff
@@ -83,10 +77,12 @@ type connector struct {
 	// use columnID instead of name since column name can be reused
 	// table names can be reused as well, deleting and adding a new table
 	// will anyway requires job restart
+	// enum mapping is only needed for enum columns with auto expand disabled for enum translation
```
voyager-dw: I think we should have Column definition, and enum mapping should go into each column level. Can keep here first if the change is too big

PR #138, opened 2019-04-17
Sharding

lxning:


client/connector.go
```diff
@@ -580,87 +347,88 @@ func (c *connector) fetchAllTables() error {
...
 }
 
-func (c *connector) fetchEnumDict(tableName, columnName string) ([]string, error) {
-	var enumDictReponse []string
-
-	resp, err := c.httpClient.Get(c.enumDictPath(tableName, columnName))
-
-	err = c.readJSONResponse(resp, err, &enumDictReponse)
+// fetchEnumDicts fetches all enum dictionaries for enum columns.
+func (c *connector) fetchEnumDicts(table metaCom.Table) (map[int]enumDict, map[int]int, error) {
```
voyager-dw: do we need to call controller? as discussed, in the future, the disabled autoexpand column should be part of the schema. So once the client get the schema, it should contain the enum already, or this is for current temporary solution?

client/connector.go
```diff
@@ -90,6 +98,13 @@ type connector struct {
 	enumDefaultValueMappings map[string]map[int]int
 }
 
+// connector is the ares connector implementation
+type connector struct {
+	*UpsertBatchBuilderImpl
```
voyager-dw: UpsertBatchBuilder interface?

client/connector.go
jshencode: this is temporary.

lxning: this is only current temporary solution. This part needs to be cleaned up once schema including enum is done.

client/connector.go
```diff
@@ -113,14 +128,17 @@ func (cfg ConnectorConfig) NewConnector(logger *zap.SugaredLogger, metricScope t
 		cfg.Timeout = defaultRequestTimeout
 	}
 
-	connector := &connector{
+	connectorCommon := UpsertBatchBuilderImpl{
```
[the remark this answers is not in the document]
lxning: will rename connectorCommon

client/connector.go
lxning: add interface UpsertBatchBuilder to separate connector from upsertBatchBuild preparation

client/connector.go
```diff
@@ -699,3 +467,168 @@ func (c *connector) dataPath(tableName string, shard int) string {
 func (c *connector) enumDictPath(tableName, columnName string) string {
 	return fmt.Sprintf("%s/%s/columns/%s/enum-cases", c.listTablesPath(), tableName, columnName)
 }
+
+// FetchTableSchema fetches table schemas from ares.
```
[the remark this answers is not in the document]
jshencode: this comment does not match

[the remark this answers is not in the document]
lxning: done

PR #139, opened 2019-04-18
move enum cases for column with disableAutoExpand into schema

jshencode:
next step
1.  add enum handler in controller
2. create enum cases in controller
3. change in disk store enum handler to get enum from schema

### DIFF
--- a/metastore/validator.go
+++ b/metastore/validator.go
@@ -101,7 +101,7 @@ func (v tableSchemaValidatorImpl) validateIndividualSchema(table *common.Table, 
 
 		// validate enum case
 		if len(column.EnumCases) > 0 {
-			if column.DisableAutoExpand {
+			if !column.DisableAutoExpand {
 				return ErrEnumCasesNotSupported
 			}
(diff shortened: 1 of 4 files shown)


metastore/validator.go
```diff
@@ -98,6 +99,23 @@ func (v tableSchemaValidatorImpl) validateIndividualSchema(table *common.Table,
...
+		if len(column.EnumCases) > 0 {
+			if column.DisableAutoExpand {
+				return ErrEnumCasesNotSupported
+			}
+			enumCaseSet := make(map[string]struct{})
+			for _, enumCase := range column.EnumCases {
+				if column.CaseInsensitive {
+					enumCase = strings.ToLower(enumCase)
+				}
+				if _, exist := enumCaseSet[enumCase]; exist {
```
voyager-dw: duplicate maybe fine, we just skip it?

jshencode: no, we should enforce this, or we will do more complicated logic to calculate enum id

PR #140, opened 2019-04-19
refactor connector

jshencode:
extract schema and enum logic
SchemaFetcher is pluggable

### DIFF
--- a/client/connector.go
+++ b/client/connector.go
@@ -16,21 +16,20 @@ package client
 
 import (
 	"bytes"
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"sync"
 	"time"
+
+	"strconv"
+	"strings"
+	"unsafe"
 
 	"github.com/uber-go/tally"
 	memCom "github.com/uber/aresdb/memstore/common"
 	metaCom "github.com/uber/aresdb/metastore/common"
 	"github.com/uber/aresdb/utils"
 	"go.uber.org/zap"
-	"strconv"
-	"strings"
-	"unsafe"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


client/schema.go
```diff
@@ -0,0 +1,405 @@
...
+	resp, err := hf.httpClient.Get(hf.tablePath(tableName))
+	err = hf.readJSONResponse(resp, err, &table)
+	if err != nil {
+		return nil, err
+	}
+
+	return &table, nil
+}
+
+func (hf *httpSchemaFetcher) FetchAllSchemas() ([]*metaCom.Table, error) {
```
voyager-dw: possible to combine into one http call?

client/schema.go
```diff
@@ -0,0 +1,405 @@
...
+		if caseInsensitive {
+			enumCase = strings.ToLower(enumCase)
+		}
+		cf.enumMappings[tableName][columnID][enumCase] = enumIDs[index]
+	}
+	cf.Unlock()
+	return nil
+}
+
+func (cf *CachedSchemaHandler) fetchAndSetEnumCases(table *metaCom.Table) error {
```
voyager-dw: also, possible to combine to fetch all column enums in one http call?

client/connector.go
```diff
@@ -527,6 +349,12 @@ func (u *UpsertBatchBuilderImpl) PrepareUpsertBatch(tableName string, columnName
 		if err = upsertBatchBuilder.AddColumnWithUpdateMode(columnID, dataType, updateModes[colIndex]); err != nil {
 			return nil, 0, err
 		}
+
+		if column.IsEnumColumn() {
+			if err = u.prepareEnumCases(tableName, columnName, colIndex, columnID, rows, abandonRows, column.CaseInsensitive, column.DisableAutoExpand); err != nil {
```
voyager-dw: this maybe an overhead for most no need to extend cases.

client/connector.go
```diff
@@ -128,43 +112,28 @@ func (cfg ConnectorConfig) NewConnector(logger *zap.SugaredLogger, metricScope t
...
-		UpsertBatchBuilderImpl: &upsertBatchBuilderImpl,
+	httpClient := http.Client{
+		Timeout: time.Duration(cfg.Timeout) * time.Second,
 	}
 
-	connector.initHTTPClient()
-
-	err := connector.fetchAllTables()
+	httpSchemaFetcher := NewHttpSchemaFetcher(httpClient, cfg.Address, metricScope)
+	cachedSchemaHandler := NewCachedSchemaHandler(logger, metricScope, httpSchemaFetcher)
```
voyager-dw: nice wrapper

client/connector.go
jshencode: this will compare with local cache first. only new cases will be extended

client/schema.go
jshencode: maybe later? currently we do not support combined call in the api. plus this is not really a performance bottleneck.

client/schema.go
jshencode: same here. maybe later when we see necessities for this change

PR #142, opened 2019-04-22
sql query one at a time

shz117:
for sql requests it makes more sense to process 1 query at a time.
this change will make it easier to stream query result batch after batch to the connection later.


shz117:
> why need to intentionally create single and multiple query context, isn't single query is a special case of multiple queries?

I think we should move away from "multiple queries per request" model. it makes it hard to stream results.

PR #143, opened 2019-04-22
Optimize ingestion

lucafuji:
Testing data: trips 8.8 GB redologs
Hosts: ares-dev
Before optimization: 3m5s 
After optimization:  5m30s


memstore/ingestion.go
```diff
@@ -363,53 +364,102 @@ func writeBatchRecords(columnDeletions []bool,
...
+			case common.UpdateWithAddition:
+				fallthrough
+			case common.UpdateWithMin:
+				fallthrough
+			case common.UpdateWithMax:
+				if !common.IsNumeric(dataType) {
+					return utils.StackError(nil, "Unsupported data type %x for column update mode %x", dataType, columnUpdateMode)
 				}
+			}
+		}
```
jshencode: actually AllValuesDefault with forceWrite = false has no need to go beyond this point right?

PR #144, opened 2019-04-23
Gateway

lxning:


gateway/controller.go
```diff
@@ -165,3 +169,65 @@ func (c *ControllerHTTPClient) GetAssignment(jobNamespace, instance string) (ass
...
+	if err != nil {
+		err = utils.StackError(err, "controller client error fetching schema for table: %s", tableName)
+		return
+	}
+	return
+}
+
+// ExtendEnumCases extends enum cases to given table column
+func (c *ControllerHTTPClient) ExtendEnumCases(tableName, columnName string, enumCases []string) (enumIDs []int, err error) {
+	enumCasesRequest := client.EnumCasesWrapper{
```
jshencode: this seems like an unnecessary dependency on client package

lxning: removed, done

client/connector.go
```diff
@@ -58,8 +58,8 @@ type UpsertBatchBuilder interface {
 	PrepareUpsertBatch(tableName string, columnNames []string, updateModes []memCom.ColumnUpdateMode, rows []Row) ([]byte, int, error)
 }
 
-// enumCasesWrapper is a response/request body which wraps enum cases
-type enumCasesWrapper struct {
+// EnumCasesWrapper is a response/request body which wraps enum cases
+type EnumCasesWrapper struct {
```
jshencode: maybe not export this struct

lxning: done

PR #145, opened 2019-04-24
support geopoint as dimension

shz117:


query/common/dimval.go
```diff
@@ -89,6 +89,14 @@ func ReadDimension(valueStart, nullStart unsafe.Pointer,
 			return &result
 		}
 		return nil
+	case memCom.GeoPoint:
+		coords := *(*[2]float32)(valuePtr)
```
lucafuji: somehow I think we can reuse the code in memcom.ConvertToHumanReadable

query/functor.hpp
```diff
@@ -575,6 +579,86 @@ struct UnaryFunctor<UUIDT, UUIDT> {
...
+  typedef thrust::tuple<GeoPointT, bool> argument_type;
+  typedef thrust::tuple<O, bool> result_type;
+
+  explicit UnaryFunctor(UnaryFunctorType functorType)
+      : functorType(functorType) {}
+
+  UnaryFunctorType functorType;
+
+  __host__ __device__ result_type operator()(const argument_type t) const {
+       O o;
```
lucafuji: Maybe a cleaner way to do this is disable binding non-geopoint output type in dimension_transform.cu

query/functor.hpp
```diff
@@ -652,11 +736,34 @@ struct BinaryFunctor<UUIDT, I> {
   }
 };
 
+template <typename I>
```
lucafuji: Why we allow binary functor for geo point?

shz117: I don't know. I didn't want to allow it.
the code won't compile without these specializations.

query/functor.hpp
shz117: sorry I didn't get it. we do still want to support other types as dimensions.

PR #152, opened 2019-04-30
fix query panic when no non-pre filter is present

shz117:
recreate to trigger travis build

### DIFF
--- a/query/aql_batchexecutor.go
+++ b/query/aql_batchexecutor.go
@@ -33,7 +33,7 @@ type BatchExecutor interface {
 	// reduce to sort and aggregate result
 	reduce()
 	// prepare work before execution
-	preExec(lastBatch bool, start time.Time, hasRegularFilter bool)
+	preExec(lastBatch bool, start time.Time)
 	// post execution after execution
 	postExec(start time.Time)
 }
@@ -59,7 +59,7 @@ func (e *DummyBatchExecutorImpl) project() {
 func (e *DummyBatchExecutorImpl) reduce() {
 }
 
-func (e *DummyBatchExecutorImpl) preExec(lastBatch bool, start time.Time, hasRegularFilter bool) {
+func (e *DummyBatchExecutorImpl) preExec(lastBatch bool, start time.Time) {
 }
 
 func (e *DummyBatchExecutorImpl) postExec(start time.Time) {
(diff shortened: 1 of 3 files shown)


query/aql_batchexecutor.go
```diff
@@ -243,10 +243,12 @@ func (e *BatchExecutorImpl) reduce() {
 	memutils.WaitForCudaStream(e.stream, e.qc.Device)
 }
 
-func (e *BatchExecutorImpl) preExec(isLastBatch bool, start time.Time) {
+func (e *BatchExecutorImpl) preExec(isLastBatch bool, start time.Time, hasRegularFilter bool) {
 	e.isLastBatch = isLastBatch
 	// initialize index vector.
-	initIndexVector(e.qc.OOPK.currentBatch.indexVectorD.getPointer(), 0, e.qc.OOPK.currentBatch.size, e.stream, e.qc.Device)
+	if hasRegularFilter {
+		initIndexVector(e.qc.OOPK.currentBatch.indexVectorD.getPointer(), 0, e.qc.OOPK.currentBatch.size, e.stream, e.qc.Device)
```
voyager-dw: would interesting to know how will it work without the index vector

shz117: for current batch, since there's no filter to evaluate, `initIndexVector`  will be skipped, `filter()` will actually do nothing.

query/aql_processor.go
```diff
@@ -725,13 +727,17 @@ func (bc *oopkBatchContext) swapResultBufferForNextBatch() {
...
-	bc.size = columns[firstColumn].length
-	bc.stats.batchSize = bc.size
-	// Allocate twice of the size to save number of allocations of temporary index vector.
-	bc.indexVectorD = deviceAllocate(bc.size*4, bc.device)
-	bc.predicateVectorD = deviceAllocate(bc.size, bc.device)
-	bc.baseCountD = columns[firstColumn].counts.offset(columns[firstColumn].countStartIndex * 4)
 	bc.startRow = startRow
+
+	if firstColumn >= 0 {
+		// some filter exists
```
jshencode: not just filter column, could be measure column or dimension column

query/aql_batchexecutor.go
```diff
@@ -243,10 +243,12 @@ func (e *BatchExecutorImpl) reduce() {
 	memutils.WaitForCudaStream(e.stream, e.qc.Device)
 }
 
-func (e *BatchExecutorImpl) preExec(isLastBatch bool, start time.Time) {
+func (e *BatchExecutorImpl) preExec(isLastBatch bool, start time.Time, hasRegularFilter bool) {
```
jshencode: hasRegularFilter parameter could be placed in BatchExecutorImpl?

query/aql_processor.go
```diff
@@ -1276,7 +1283,7 @@ func (qc *AQLQueryContext) FindDeviceForQuery(memStore memstore.MemStore, prefer
 
 func (qc *AQLQueryContext) runBatchExecutor(e BatchExecutor, isLastBatch bool) {
 	start := utils.Now()
-	e.preExec(isLastBatch, start)
+	e.preExec(isLastBatch, start, !qc.OOPK.currentBatch.indexVectorD.isNull())
```
jshencode: looks like param hasRegularFilter is redundent, we can just check indexVectorD.isNull() directly before we initialize indexVector

shz117: true.. somehow I thought indexVectorD is not available from batchExecutor.

PR #154, opened 2019-05-01
fix pre filter problem

lucafuji:
when slideIndex, the last range of upperbound/lowerbound should be vp.length + 1

### DIFF
--- a/memstore/vector_party.go
+++ b/memstore/vector_party.go
@@ -412,8 +412,12 @@ func (vp *cVectorParty) SliceByValue(lowerBoundRow, upperBoundRow int, value uns
 // party.
 func (vp *cVectorParty) SliceIndex(lowerBoundRow, upperBoundRow int) (startIndex, endIndex int) {
 	if vp.GetMode() == common.HasCountVector {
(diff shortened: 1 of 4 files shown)


memstore/vector_party.go
```diff
@@ -412,8 +412,12 @@ func (vp *cVectorParty) SliceByValue(lowerBoundRow, upperBoundRow int, value uns
 // party.
 func (vp *cVectorParty) SliceIndex(lowerBoundRow, upperBoundRow int) (startIndex, endIndex int) {
 	if vp.GetMode() == common.HasCountVector {
-		startIndex = vp.counts.UpperBound(0, vp.length, unsafe.Pointer(&lowerBoundRow)) - 1
-		endIndex = vp.counts.LowerBound(startIndex, vp.length, unsafe.Pointer(&upperBoundRow))
+		startIndex = vp.counts.UpperBound(0, vp.length+1, unsafe.Pointer(&lowerBoundRow))
```
shz117: why `vp.length+1` not `vp.length`..?

lucafuji: because counts vector is one element longer

PR #156, opened 2019-05-01
kafka redolog manager

jshencode:


memstore/kafka_redolog_manager.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+	}()
+
+	return nil
+}
+
+func (k *kafkaRedologManager) Close() {
+	if k.partitionConsumer != nil {
+		k.partitionConsumer.Close()
```
shz117: if we don't set nil here, would it panic at line 118?

memstore/kafka_redolog_manager.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"encoding/json"
+	"github.com/Shopify/sarama"
+	"github.com/uber/aresdb/utils"
+	"math"
+	"sync"
+)
+
+const maxBatchesPerFile = 5000
```
shz117: should this be configurable?

memstore/kafka_redolog_manager.go
```diff
@@ -0,0 +1,168 @@
...
+				return batchBundle.batch, fileID, uint32(fileOffset)
+			case <-k.done:
+				// redolog manager closed
+				return nil, 0, 0
+			}
+		}
+	}
+}
+
+func (k *kafkaRedologManager) PurgeRedologFileAndData(eventTimeCutoff uint32, fileIDCheckpointed int64, batchOffset uint32) error {
```
shz117: it doesn't purge data in this implementation, maybe add a comment, if renaming interface is tricky

memstore/kafka_redolog_manager.go
```diff
@@ -0,0 +1,168 @@
...
+			// file not purgeable
+			if fileID < firstUnpurgeable {
+				firstUnpurgeable = fileID
+			}
+		}
+	}
+	k.RUnlock()
+
+	if firstUnpurgeable < math.MaxInt64 {
+		err := k.commitFunc(firstUnpurgeable * maxBatchesPerFile)
```
voyager-dw: this is not an valid kafka offset, maybe add 2nd param for batch offset?

memstore/kafka_redolog_manager.go
```diff
@@ -0,0 +1,168 @@
...
+	k.partitionConsumer, err = k.consumer.ConsumePartition(k.Topic, int32(k.Shard), offset)
+	if err != nil {
+		return utils.StackError(err, "failed to consume from topic %s partition %d",
+			k.Topic, k.Shard)
+	}
+
+	go func() {
+		for {
+			select {
+			case msg, ok := <-k.partitionConsumer.Messages():
```
voyager-dw: how about the error channel?

memstore/kafka_redolog_manager.go
jshencode: if user close redolog manager, they should not start consumer from it

memstore/kafka_redolog_manager.go
jshencode: for now, this number should be fixed otherwise saved progress (consist of fileID and offset) is not useable after this number is changed.

memstore/kafka_redolog_manager.go
jshencode: change to CheckpointRedolog

memstore/kafka_redolog_manager.go
jshencode: good point, adding FirstKafkaOffsetPerFile, should always checkpoint the first kafka offset per file

memstore/kafka_redolog_manager.go
jshencode: consuming from error channel is optional. sarama will log the error even we do not consume from it. but we can add the error channel as well

subscriber/config/test-ares-ns-empty.yaml
```diff
@@ -7,11 +7,13 @@ ares:
   clusters:
     clusters:
```
voyager-dw: instances?

jshencode: not belong to this diff

PR #157, opened 2019-05-01
table incarnation

jshencode:


metastore/schema_fetch.go
```diff
@@ -99,7 +99,20 @@ func (j *SchemaFetchJob) applySchemaChange(tables []common.Table) (err error) {
 			if err != nil {
 				return
 			}
-			if !reflect.DeepEqual(&table, oldTable) {
+			if oldTable.Incarnation < table.Incarnation {
```
voyager-dw: looks good for new code, line 89 better updated to check map contains

PR #159, opened 2019-05-03
refactor non agg query execution

shz117:
After this PR we should be able to remove non-agg query result size limit from engine level.
 - refactor post processor: extract flushContext and flush function
 - non-agg query executor will flush result after each batch processing
 - update non-agg query memory estimation


query/aql_nonaggr_batchexecutor.go
```diff
@@ -39,53 +40,57 @@ func (e *NonAggrBatchExecutorImpl) project() {
...
-	if bc.size > lenWanted {
-		bc.size = lenWanted
+}
+
+func (e *NonAggrBatchExecutorImpl) expandDimensions(numDims queryCom.DimCountsPerDimWidth) {
+	bc := &e.qc.OOPK.currentBatch
+
+	if bc.size != 0 && !bc.baseCountD.isNull() {
+		e.qc.doProfile(func() {
+			e.qc.OOPK.currentBatch.expand(numDims, bc.size, e.stream, e.qc.Device)
```
jshencode: expand will expand dimensionVectorD[0] into dimensionVectorD[1]
and in postExec, L77 it actually writes dimensionVectorD results into dimensionVectorH, which is uncompressed dim result.

we might need to swap dimensionVectorD[0] and dimensionVectorD[1] in expand

query/aql_nonaggr_batchexecutor.go
```diff
@@ -39,53 +40,57 @@ func (e *NonAggrBatchExecutorImpl) project() {
...
 
 	bc := &e.qc.OOPK.currentBatch
+	lenWanted := e.qc.Query.Limit - e.qc.numberOfRowsWritten
+	if bc.size > lenWanted {
+		bc.size = lenWanted
+	}
+
 	if bc.resultCapacity == 0 {
-		bc.resultCapacity = e.qc.Query.Limit
+		bc.resultCapacity = bc.size
```
jshencode: in this case, if we don't have a limit for the query. resultCapacity should be the maximum size of all batches, not the first batch size

shz117: I think the buffer capacity should be the size of the current batch after pre filtering. 
to reduce confusion, maybe I should not reuse the `resultCapacity` field.

query/aql_nonaggr_batchexecutor.go
shz117: right, good catch

query/aql_nonaggr_batchexecutor.go
shz117: actually you are right, since we allocate buffer for current and next.

voyager-dw: +1

query/aql_processor.go
```diff
@@ -105,6 +105,17 @@ func (qc *AQLQueryContext) ProcessQuery(memStore memstore.MemStore) {
...
+	if qc.isNonAggregationQuery {
+		qc.Results = make(queryCom.AQLQueryResult)
+		headers := make([]string, len(qc.Query.Dimensions))
+		for i, dim := range qc.Query.Dimensions {
+			headers[i] = dim.Expr
+		}
+		qc.Results.SetHeaders(headers)
+	}
+
+	qc.initResultFlushContext()
```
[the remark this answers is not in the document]
shz117: yes

query/aql_processor.go
```diff
@@ -1127,10 +1143,21 @@ func (qc *AQLQueryContext) estimateMemUsageForBatch(firstColumnSize int, columnM
...
+	}
 
 	// 8. Dimension vector memory usage (input + output)
-	memUsage += firstColumnSize * qc.OOPK.DimRowBytes * 2
+	if qc.isNonAggregationQuery {
+		maxRowsPerBatch := maxSizeAfterPreFilter
+		if qc.Query.Limit < maxRowsPerBatch {
+			maxRowsPerBatch = qc.Query.Limit
+		}
+		memUsage += maxRowsPerBatch * qc.OOPK.DimRowBytes * 2
```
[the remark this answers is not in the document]
shz117: maxSizeAfterPreFilter tracks the number of rows if uncompressed, so the same calculation applies whether it's sorted or non sorted column

shz117:
> instead of flushing for every batch, maybe better use lazy on demand flushing after we find out the new batch result can not fit into existing buffer?

that's a good idea. however initial buffer size and memory estimation can be tricky.
this can be an optimization to do for next step.

PR #167, opened 2019-05-14
integrate with fossa

lucafuji:


lucafuji:
@xinding33 I created another PR and tried to call fossa, and we got following error
```
FATAL Could not analyze: open thirdparty/rmm/python/requirements.txt: no such file or directory
```
it seems just the library rmm that we depend on don't have the equirements.txt, is there any way to exclude this directory so our build will not fail?

PR #170, opened 2019-05-15
Issue 169

lxning:


gateway/controller.go
```diff
@@ -178,8 +178,16 @@ func (c *ControllerHTTPClient) SetNamespace(namespace string) {
 }
 
 // FetchAllSchemas fetches all schemas
-func (c *ControllerHTTPClient) FetchAllSchemas() (tables []common.Table, err error) {
-	return c.GetAllSchema(c.namespace)
+func (c *ControllerHTTPClient) FetchAllSchemas() (tables []*common.Table, err error) {
+	schemas, err := c.GetAllSchema(c.namespace)
```
lucafuji: here since you are using :=, the err will shadow the err in the return arguments, so anyway you will return a nil err.
I recommend you either using non-named return arguments or pre declare schemas

lucafuji: BTW, I think you need to write a test case to catch above non-nil error even it fails

gateway/controller_test.go
```diff
@@ -155,9 +158,9 @@ var _ = ginkgo.Describe("Controller", func() {
 		Ω(err).Should(BeNil())
 
 		c.SetNamespace("ns1")
-		tablesGot, err = c.FetchAllSchemas()
+		tableAddressesGot, err := c.FetchAllSchemas()
 		Ω(err).Should(BeNil())
-		Ω(tablesGot).Should(Equal(tables))
+		Ω(tableAddressesGot).Should(Equal(tableAddresses))
```
lucafuji: better to add another test case with two table

lucafuji:
BTW give the pull request a better name ...

gateway/controller.go
lxning: good catch

gateway/controller_test.go
lxning: done

PR #171, opened 2019-05-15
table shard level option to disable disk purge in archiving, backfill, snapshot process

jshencode:


memstore/table_shard.go
```diff
@@ -71,19 +71,27 @@ func NewTableShard(schema *TableSchema, metaStore metastore.MetaStore,
...
+
+// DoneWithPeerDataTransfer unpin disk file after data transfer is done
+func (shard *TableShard) DoneWithPeerDataTransfer() {
+	pins := atomic.AddInt32(&shard.pinnedDiskFileUsers, -1)
+	utils.GetReporter(shard.Schema.Schema.Name, shard.ShardID).GetGauge(utils.PinnedDiskFileUsers).
+		Update(float64(pins))
 }
 
 // IsDiskPurgeEnabled check whether disk purge is enabled
 func (shard *TableShard) IsDiskPurgeEnabled() bool {
```
voyager-dw: how will this function be used? The check may be incorrect if it is enabled/disabled after this function call

jshencode: This function will be used before archiving disk purge step

PR #173, opened 2019-05-21
eager flush results for single non aggregate query

shz117:
this will reduce TTFB for non-agg queries from 10+s to milliseconds.

### DIFF
--- a/api/query_handler.go
+++ b/api/query_handler.go
@@ -90,7 +90,7 @@ func (handler *QueryHandler) handleAQLInternal(aqlRequest AQLRequest, w http.Res
 	var qcs []*query.AQLQueryContext
 	var statusCode int
 
-	defer func() {
+	defer func () {
 		var errStr string
 		if err != nil {
 			errStr = err.Error()
@@ -111,7 +111,6 @@ func (handler *QueryHandler) handleAQLInternal(aqlRequest AQLRequest, w http.Res
 		} else {
 			l.Error("Some of the queries finished with error")
 		}
-
 	}()
 
 	if aqlRequest.Query != "" {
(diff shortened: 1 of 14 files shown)


query/aql_compiler.go
```diff
@@ -79,8 +80,8 @@ const (
 
 // Compile returns the compiled AQLQueryContext for data feeding and query
 // execution. Caller should check for AQLQueryContext.Error.
-func (q *AQLQuery) Compile(store memstore.MemStore, returnHLL bool) *AQLQueryContext {
-	qc := &AQLQueryContext{Query: q, ReturnHLLData: returnHLL}
+func (q *AQLQuery) Compile(store memstore.MemStore, returnHLL bool, w http.ResponseWriter) *AQLQueryContext {
```
voyager-dw: it is quite strange compile needs the http writer as parameter, maybe outside assign to aqlquerycontext?

api/query_handler.go
```diff
@@ -138,37 +137,88 @@ func (handler *QueryHandler) handleAQLInternal(aqlRequest AQLRequest, w http.Res
...
+			err = qc.Error
+			utils.GetQueryLogger().With(
+				"error", err,
+				"query", aqlQuery,
+				"context", qc,
+			).Error("Error happened when processing query")
+			statusCode = http.StatusInternalServerError
+			return
+		}
+		w.Write([]byte(`]}]`))
```
voyager-dw: where is the [{[, inside processquery?

shz117: yes..

query/aql_compiler.go
shz117: good point, I can change it

PR #174, opened 2019-05-21
recover from job panic

shz117:
so it doesn't crash the server.

### DIFF
--- a/memstore/scheduler.go
+++ b/memstore/scheduler.go
@@ -231,7 +231,7 @@ func (scheduler *schedulerImpl) executeJob(jb *jobBundle) {
 	defer func() {
 		if r := recover(); r != nil {
 			utils.GetRootReporter().GetCounter(utils.JobFailuresCount).Inc(1)
-			utils.GetLogger().With("error", r).Error("Job failed after panic")
+			utils.GetLogger().With("error", r, "job", jb.Job).Error("Job failed after panic")
 		}
 	}()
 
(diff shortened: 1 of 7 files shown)


memstore/scheduler.go
```diff
@@ -228,6 +228,12 @@ func (scheduler *schedulerImpl) Start() {
 }
 
 func (scheduler *schedulerImpl) executeJob(jb *jobBundle) {
+	defer func() {
```
lucafuji: IMO better to crash the server to catch the bug

memstore/scheduler.go
```diff
@@ -228,6 +228,12 @@ func (scheduler *schedulerImpl) Start() {
...
+		if r := recover(); r != nil {
+			utils.GetLogger().Errorf("Job failed a fter panic: %v", r)
```
lucafuji: use log fields and there is a typo

lucafuji: And better to treat the error differently according to its type

memstore/scheduler.go
shz117: crashing server means downtime, can't be good?

we can add metrics to discover job failure, plus logs to investigate.

memstore/scheduler.go
```diff
@@ -228,8 +228,15 @@ func (scheduler *schedulerImpl) Start() {
 }
 
 func (scheduler *schedulerImpl) executeJob(jb *jobBundle) {
+	defer func() {
+		if r := recover(); r != nil {
+			utils.GetRootReporter().GetCounter(utils.JobFailuresCount).Inc(1)
+			utils.GetLogger().With("error", r).Error("Job failed after panic")
```
lucafuji: better to have table, shard and job type information

PR #182, opened 2019-05-22
update enum path, log and unit test

lxning:


shz117:
it's more like "add logging and metrics for etcd integration"..
can you rename the PR or add some context in the description the help ppl review?

subscriber/common/job/controller.go
```diff
@@ -143,27 +143,29 @@ func NewController(params Params) *Controller {
 		params.ServiceConfig.Logger.Info("aresDB Controller is enabled")
 
 		if params.ServiceConfig.HeartbeatConfig.Enabled {
+			params.ServiceConfig.Logger.Info("heartbeat config",
+				zap.Any("heartbeat", *params.ServiceConfig.HeartbeatConfig))
 			controller.etcdServices, err = connectEtcdServices(params)
 			if err != nil {
 				panic(utils.StackError(err, "Failed to createEtcdServices"))
```
jshencode: better use logger.fatal so we can catch the log in elk

subscriber/common/job/controller.go
```diff
@@ -143,27 +143,29 @@ func NewController(params Params) *Controller {
...
-		err = controller.zkClient.Start()
-		if err != nil {
-			panic(utils.StackError(err, "Failed to start zkClient"))
-		}
-		params.ServiceConfig.Logger.Info("zkClient was started")
+		} else {
+			controller.zkClient = createZKClient(params)
+			err = controller.zkClient.Start()
+			if err != nil {
+				panic(utils.StackError(err, "Failed to start zkClient"))
```
jshencode: same here

subscriber/common/job/controller.go
```diff
@@ -143,27 +143,29 @@ func NewController(params Params) *Controller {
...
+			}
+			params.ServiceConfig.Logger.Info("zkClient was started")
 
-		err = controller.RegisterOnZK()
-		if err != nil {
-			panic(utils.StackError(err, "Failed to register subscriber"))
+			err = controller.RegisterOnZK()
+			if err != nil {
+				panic(utils.StackError(err, "Failed to register subscriber"))
```
jshencode: same here

subscriber/common/job/controller.go
lxning: here should stop the service to alert deployment failed via panic.

subscriber/common/job/controller.go
lxning: same here.

subscriber/common/job/controller.go
```diff
@@ -18,13 +18,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/m3db/m3/src/cluster/placement"
```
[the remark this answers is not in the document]
lxning: yes, from auto fmt.

subscriber/common/job/controller.go
lxning: replace with logger.panic

subscriber/common/job/controller.go
lxning: same

subscriber/common/job/controller.go
lxning: same

PR #184, opened 2019-05-24
include cudf 

lucafuji:


shz117:
care to explain how cudf will be used?

lucafuji:
> care to explain how cudf will be used?

Will use it to do hash reduction and future hash join

PR #188, opened 2019-05-29
peer data streaming grpc

jshencode:
will rebase after go.mod diff is merged to fix the apt-get issue


datanode/generated/proto/rpc/peer_streaming.proto
```diff
@@ -0,0 +1,117 @@
...
+    }
+    repeated BatchMetaData batches = 7; // batches
+}
+
+message TableShardMetaDataRequest {
+    string table = 1; // table name
+    int32  incarnation = 2; // table incarnation
+    uint32 shard = 3; // shard id
+
+    int32 startBatchID = 4; // first batch id to fetch, fact only
```
voyager-dw: start/End batchID should be in the response instead of request, no?

.travis/install_clang_tools.sh
```diff
@@ -4,7 +4,6 @@ set -ex
 output_dir=./build
 cmake_version=3.12.0
 # install cppcheck
-sudo apt-get update
```
lucafuji: this is not a good way to resolve this, maybe better to rebase my diff to solve the problem

datanode/generated/proto/rpc/peer_streaming.proto
jshencode: request can specify a range of batches to fetch.
response will return a list of batch metadata

datanode/generated/proto/rpc/peer_streaming.proto
```diff
@@ -0,0 +1,117 @@
...
+}
+
+message BackfillCheckpoint {
+    int64 redoFileID = 1; // checkpoint redolog file id, dimension table only
+    uint32 redoFileOffset = 2; // checkpoint redolog file offset, dimension table only
+}
+
+message ArchiveVersion {
+    uint32 archiveVersion = 1; // archive version, fact table only
+    uint32 backfillSeq = 2; // backfillSeq, fact table only
```
shz117: what is this..?

datanode/generated/proto/rpc/peer_streaming.proto
```diff
@@ -0,0 +1,117 @@
...
+message Session {
+    int64 ID = 1;
+}
+
+message KeepAliveResponse {
+    int64 ID = 1;
+    int64 ttl = 2;
+}
+
+service PeerDataNode {
```
shz117: some comments/documentation about how the interaction works may help (e.g. sequence of calling endpoints below).
it could be added here or comments of some other code, or a dedicated readme.

datanode/generated/proto/rpc/peer_streaming.proto
jshencode: each backfill will increment a sequence number, which is part of the archive batch's version

datanode/generated/proto/rpc/peer_streaming.proto
jshencode: added comments

PR #191, opened 2019-05-31
fix controller ingestion assignment bug and tests random failures

jshencode:
1.  remove etcd embed server 
2. use gomock and in memory kv.Store implementation for testing
3. fix ingestion assignment bug: add waiting for leader election result before actual assignment loop starts


controller/tasks/etcd/ingestion_assignment.go
```diff
@@ -173,33 +173,57 @@ func (ia *ingestionAssignmentTask) Done() {
...
+			}
+		case <-ia.stopChan:
+			return
+		}
+
+		ia.logger.With("host", hostName).Infof("elected as leader")
+		namespaces, err := ia.namespaceMutator.ListNamespaces()
 		if err != nil {
-			ia.logger.Fatal(err)
+			ia.logger.Error(err)
```
shz117: should this be fatal as well?

controller/tasks/etcd/ingestion_assignment.go
```diff
@@ -173,33 +173,57 @@ func (ia *ingestionAssignmentTask) Done() {
...
-		case <-tickerChan:
-			if ia.isLeader() {
+			// periodic updates
+			case <-tickerChan:
+				if !ia.isLeader() {
+					ia.logger.With("host", hostName).Infof("host is no longer the leader")
+					break loop
+				}
 				ia.tryRecalculateAllNamespaces()
+			// periodic updates
```
shz117: nit

controller/tasks/etcd/ingestion_assignment.go
```diff
@@ -173,33 +173,57 @@ func (ia *ingestionAssignmentTask) Done() {
...
-		select {
-		// watch changes (i.e: heartbeat changes)
-		case ns := <-ia.watchManager.C():
-			if ia.isLeader() {
+		tickerChan := time.NewTicker(time.Duration(ia.intervalSeconds) * time.Second).C
+
+	loop:
+		for {
+			select {
+			case <- ia.leaderElection.C():
```
shz117: do we need this? this process should already be leader past line 180, right?

shz117: oh, is it for resign?

jshencode: yes, this is for when resign happens

controller/tasks/etcd/ingestion_assignment.go
jshencode: done

controller/tasks/etcd/ingestion_assignment.go
jshencode: done

PR #192, opened 2019-05-31
Broker skeleton

shz117:
changes included:
 - minor refactoring to prepare for broker implementation: memstore + aql context
 - skeleton code for broker

TODOs are added including:
 - implement BrokerSchemaMutator
 - implement non agg query executor
 - implement agg query executor
 - implement different query options in handler
 - logging and metrics where needed


datanode/topology/ares_suite_test.go
```diff
@@ -0,0 +1,16 @@
+package topology
```
shz117: topology pkg might be moved or subject to changes.

broker/broker_schema_mutator.go
```diff
@@ -0,0 +1,69 @@
+package broker
+
+import (
+	"github.com/uber/aresdb/metastore/common"
+	"github.com/uber/aresdb/metastore"
+	"github.com/uber/aresdb/memstore"
```
jshencode: depending on memstore will bring in unnecessary dependency
please rebase from newest change, david has moved memstore.TableSchema to common package

broker/executor.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// QueryExecutor defines query executor
+type QueryExecutor interface {
+	// Execute executes query and flush result to connection
+	Execute(namespace, sqlQuery string, w http.ResponseWriter) (err error)
+}
+
+// NewQueryExecutor creates a new QueryExecutor
+func NewQueryExecutor(sm SchemaManager, topo topology.DynamicTopology) QueryExecutor {
```
jshencode: since we do not need to MarkShardsAvailable in broker, we can just use topology.Topology instead of DynamicTopology

broker/handler.go
```diff
@@ -0,0 +1,70 @@
...
+	exec QueryExecutor
+}
+
+func NewQueryHandler(executor QueryExecutor) QueryHandler {
+	return QueryHandler{
+		exec: executor,
+	}
+}
+
+func (handler *QueryHandler)Register(router *mux.Router, wrappers ...utils.HTTPHandlerWrapper) {
```
jshencode: nit: format

cmd/broker/cmd/cmd.go
```diff
@@ -0,0 +1,114 @@
...
+
+	for _, setter := range setters {
+		setter(options)
+	}
+
+	cmd := &cobra.Command{
+		Use:     "aresbrokerd",
+		Short:   "AresDB broker",
+		Long:    `AresDB broker is the gateway to send queries to AresDB`,
+		Example: `./ares --config config/ares-broker.yaml --port 9374`,
```
jshencode: nit: broker also called ares?

shz117: oops, will change

broker/broker_schema_mutator.go
voyager-dw: +1 broker likely should only need to know schema and topology

broker/broker_schema_mutator.go
```diff
@@ -0,0 +1,69 @@
...
+func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err error) {
+	tableSchema, ok := b.tables[name]
+	if !ok {
+		err = metastore.ErrTableDoesNotExist
+	}
+	table = &tableSchema.Schema
+	return
+}
+
+func (b *BrokerSchemaMutator) CreateTable(table *common.Table) (err error) {
```
voyager-dw: since broker only get schema sync from the controller, we may not need to expose these functions

memstore/memstore.go
```diff
@@ -31,19 +31,27 @@ type TableShardMemoryUsage struct {
 	PrimaryKeyMemory uint                                 `json:"pk"`
 }
 
+type TableSchemaReader interface {
```
jshencode: might want to move this interface to common, so broker will not depend on memstore

cmd/broker/cmd/cmd.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/uber/aresdb/broker/config"
+	"github.com/uber/aresdb/cmd/aresd/cmd"
+	"github.com/uber/aresdb/common"
+	"github.com/uber/aresdb/cluster/topology"
+	"github.com/uber/aresdb/controller/client"
+	"github.com/uber/aresdb/utils"
+	"time"
+)
+
+func Execute(setters ...cmd.Option) {
```
voyager-dw: better to use fx instead of migration later

broker/broker_schema_mutator.go
shz117: this is needed as BrokerSchemaMutator struct implements TableSchemaMutator interface and mutates an in-memory map of tables.

cmd/broker/cmd/cmd.go
shz117: not sure I get it..  do you mean using array?

voyager-dw:
can we remove those duplicated stuff or create a new pull request?

shz117:
> can we remove those duplicated stuff or create a new pull request?

sure I will rebase on master since the cluster pkg is already checked in

cmd/broker/cmd/cmd.go
voyager-dw: it is fine, as long as from internal package we can use uberfx to call

memstore/memstore.go
shz117: sg. will move some interface def in metastore to common as well

broker/schema_manager.go
```diff
@@ -0,0 +1,75 @@
...
+		job := metastore.NewSchemaFetchJob(schemaFetchInterval, mutator, metastore.NewTableSchameValidator(), sm.controllerCli, namespace, "")
+		job.FetchSchema()
+		go job.Run()
+		sm.mutators[namespace] = mutator
+	}
+}
+
+func (sm *schemaManagerImpl) GetTableSchemaReader(namespace string) (tableSchemaReader memstore.TableSchemaReader, err error) {
+	var exists bool
+	tableSchemaReader, exists = sm.mutators[namespace]
```
voyager-dw: mutator or reader? or from same interface?

broker/schema_manager.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Run initializes all schema and starts jobs to sync from controller
+func (sm *schemaManagerImpl) Run() {
+	namespaces, err := sm.controllerCli.GetNamespaces()
+	if err != nil {
+		utils.GetLogger().Fatal("Failed to fetch namespaces from controller", err)
+	}
+
+	for _, namespace := range namespaces {
```
voyager-dw: how about new name space?

broker/executor.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+}
+
+// queryExecutorImpl will be reused across all queries
+type queryExecutorImpl struct {
+	schemaManager SchemaManager
+	topo          topology.Topology
+}
+
+func (qe *queryExecutorImpl) Execute(namespace, sqlQuery string, w http.ResponseWriter) (err error) {
```
voyager-dw: namespace is part of the table, or separate from query?

broker/schema_manager.go
```diff
@@ -0,0 +1,75 @@
...
+	Run()
+	// GetTable gets schema by namespace and table name
+	GetTableSchemaReader(namespace string) (memstore.TableSchemaReader, error)
+}
+
+type schemaManagerImpl struct {
+	sync.RWMutex
+
+	controllerCli client.ControllerClient
+	mutators      map[string]*BrokerSchemaMutator
```
voyager-dw: or do we really need mutator?

broker/schema_manager.go
shz117: this function will be called to get a TableSchemaReader and pass to compiler, it doesn't matter if it's also a mutator, thus the name

broker/schema_manager.go
shz117: ah right. will need another pulling job to get new namespaces.

broker/executor.go
shz117: I think it can be separate from query. because cross namespace join is not supported anyway.

broker/schema_manager.go
shz117: yeah it's needed by SchemaFetchJob

broker/schema_manager.go
shz117: added todo item. didn't want this diff to bloat.

PR #195, opened 2019-06-04
fix non agg query

shz117:
- fix sql parser populating LIMIT
 - fix result flushing

### DIFF
--- a/query/sql_parser.go
+++ b/query/sql_parser.go
@@ -354,12 +354,18 @@ func (v *ASTBuilder) VisitQueryNoWith(ctx *antlrgen.QueryNoWithContext) interfac
 	v.SQL2AqlCtx.exprOrigin = ExprOriginOthers
 	var orderBy = v.getOrderBy(ctx)
 
+	limit, err := strconv.Atoi(v.GetTextIfPresent(ctx.GetLimit()))
+	if err != nil {
+		v.Logger.Panic("failed to parse limit")
+	}
+
 	var query *tree.Query
 	if qSpec, ok := term.(*tree.QuerySpecification); ok {
 		qSpecNew := tree.NewQuerySpecification(v.getLocation(ctx),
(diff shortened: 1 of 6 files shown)


query/sql_parser.go
```diff
@@ -354,9 +354,11 @@ func (v *ASTBuilder) VisitQueryNoWith(ctx *antlrgen.QueryNoWithContext) interfac
 	v.SQL2AqlCtx.exprOrigin = ExprOriginOthers
 	var orderBy = v.getOrderBy(ctx)
 
-	limit, err := strconv.Atoi(v.GetTextIfPresent(ctx.GetLimit()))
-	if err != nil {
-		v.Logger.Panic("failed to parse limit")
+	limit := 0
+	if limitParsed, err := strconv.Atoi(v.GetTextIfPresent(ctx.GetLimit())); err != nil {
+		v.Logger.Warn("failed to parse limit")
```
voyager-dw: this should be a parse error and return

shz117: yeah.. I've updated the panic msg to contain more context. the panic will be recovered and throw to handler as error.

shz117:
looks like the coverage diff is not correct, will merge..

query/sql_parser.go
```diff
@@ -354,12 +354,22 @@ func (v *ASTBuilder) VisitQueryNoWith(ctx *antlrgen.QueryNoWithContext) interfac
 	v.SQL2AqlCtx.exprOrigin = ExprOriginOthers
 	var orderBy = v.getOrderBy(ctx)
 
+	limit := 0
+	limitNode := ctx.GetLimit()
+	if limitParsed, err := strconv.Atoi(v.GetTextIfPresent(limitNode)); limitNode != nil && err != nil {
+		limitLoc := v.getLocation(limitNode)
+		v.Logger.Panicf("failed to parse limit %s at (line:%d, col:%d)", limitNode.GetText(), limitLoc.Line, limitLoc.CharPosition)
```
voyager-dw: oh no error to return here? panic will be too bad, or recover outside?

PR #197, opened 2019-06-06
add merge node

shz117:
- refactor interfaces to common pkg
 - add query plan, define plan nodes
 - implement resultMerge logic

### DIFF
--- a/broker/query_plan.go
+++ b/broker/query_plan.go
@@ -1,15 +1,27 @@
+//  Copyright (c) 2017-2018 Uber Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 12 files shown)


broker/query_plan.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/uber/aresdb/broker/common"
+	"github.com/uber/aresdb/cluster/topology"
+	"github.com/uber/aresdb/query"
+	queryCom "github.com/uber/aresdb/query/common"
+	"github.com/uber/aresdb/utils"
+	"sync"
+)
+
+type blockingPlanNodeImpl struct {
+	children []common.BlockingPlanNode
```
voyager-dw: something to remind:
1. each shard will have multiple owners
2. within query timeout, if one request to one shard owner fail, can retry other shard owner
3. hll need to support two mode, we can disscuss

shz117: for 1 and 2. RPC calls will be implemented in a different BlockingPlanNode (RPCNode), each of which will own a shard for this query, and handle retry within itself. any error returned from RPCNode will be considered un-recoverable. timeout can be passed to RPCNode as Context.

PR #200, opened 2019-06-10
[Broker] add ScanNode

shz117:
- implement ScanNode which fetches data from datanode


broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
+		hosts, routeErr := sn.topo.Get().RouteShard(sn.shardID)
+		if routeErr != nil {
+			utils.GetLogger().With(
+				"error", routeErr,
+				"shard", sn.shardID,
+				"trial", trial).Error("failed to route shard")
+			err = utils.StackError(routeErr, "route shard failed")
+			continue
+		}
+		// pick random routable host
```
voyager-dw: avoid the one already failed?

shz117: I thought about that, but seems unnessasary to remember the last host. the assumption here is that any unhealthy hosts will be removed from topology, then all hosts that are still in topology should be equal.

voyager-dw: fine for now, better we add the logic in the future, topology just use heartbeat, sometime error may due to other causes

broker/common/types.go
```diff
@@ -32,16 +33,21 @@ type SchemaManager interface {
...
 type QueryExecutor interface {
 	// Execute executes query and flush result to connection
-	Execute(namespace, sqlQuery string, w http.ResponseWriter) (err error)
+	Execute(ctx context.Context, namespace, sqlQuery string, w http.ResponseWriter) (err error)
 }
 
 // BlockingPlanNode defines query plan nodes that waits for children to finish
 type BlockingPlanNode interface {
-	Run() (queryCom.AQLQueryResult, error)
+	Run(ctx context.Context) (queryCom.AQLQueryResult, error)
```
voyager-dw: maybe all change to use Execute() for consistency?

broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
+		return
+	}
+
 	result = childrenResult[0]
 	for i := 1; i < nChildren; i++ {
-		ctx := newResultMergeContext(mn.AggType)
-		result = ctx.run(result, childrenResult[i])
-		if ctx.err != nil {
-			err = ctx.err
+		mergeCtx := newResultMergeContext(mn.aggType)
```
voyager-dw: future improve: early merge before waiting for all result come back

query/expr/ast.go
```diff
@@ -53,6 +53,26 @@ var typeNames = map[Type]string{
...
+	DayOfWeekCallName           = "dayofweek"
+	FromUnixTimeCallName        = "from_unixtime"
+	GeographyIntersectsCallName = "geography_intersects"
+	HexCallName                 = "hex"
+	// hll aggregation function applies to hll columns
+	HllCallName = "hll"
+	// countdistincthll aggregation function applies to all columns, hll value is computed on the fly
+	CountDistinctHllCallName = "countdistincthll"
+	HourCallName             = "hour"
+	ListCallName             = ""
```
voyager-dw: empty?

broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
-		if common.Count != rhs.AggType {
+		if common.Count != rhs.AggType() {
 			err = utils.StackError(nil, "RHS of avg node must be count node")
 			return
 		}
 	}
 
 	childrenResult := make([]queryCom.AQLQueryResult, nChildren)
-	errs := make([]error, nChildren)
+	errs := 0
```
lucafuji: better name it as nerrs, errs sounds like an array of err not a count

broker/query_plan.go
```diff
@@ -138,5 +222,43 @@ func (ap *AggQueryPlan) Run() (results queryCom.AQLQueryResult, err error) {
 //3. fan out requests, upon data from data nodes, flush to w
 //4. close, clean up, logging, metrics
 // TODO
-type NonAggQueryPlan struct {
+type NonAggQueryPlan struct{}
+
+// splitAvgQuery to sum and count queries
+func splitAvgQuery(q query.AQLQuery) (sumq query.AQLQuery, countq query.AQLQuery) {
```
lucafuji: since we already support avg aggregation in a singlr query, not sure why you need to split

broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
+	for trial < rpcRetries {
+		trial++
 
+		hosts, routeErr := sn.topo.Get().RouteShard(sn.shardID)
+		if routeErr != nil {
+			utils.GetLogger().With(
+				"error", routeErr,
+				"shard", sn.shardID,
+				"trial", trial).Error("failed to route shard")
+			err = utils.StackError(routeErr, "route shard failed")
```
lucafuji: nit: use same error message for error and log

broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
+			continue
+		}
+		// pick random routable host
+		rand.Seed(time.Now().UTC().UnixNano())
```
lucafuji: does seem to need to seed every time for each trial

broker/query_plan.go
```diff
@@ -35,75 +46,138 @@ func (bpn *blockingPlanNodeImpl) Add(nodes ...common.BlockingPlanNode) {
...
+				"trial", trial).Error("failed to fetch from datanode")
+			err = utils.StackError(fetchErr, "fetch from datanode failed")
+			continue
+		}
+		utils.GetLogger().With(
+			"trial", trial,
+			"shard", sn.shardID,
+			"host", host).Info("fetch from datanode succeeded")
+	}
+	if result != nil {
```
lucafuji: under what circumstance will a result be non nil and err be nil?

broker/query_plan.go
```diff
@@ -112,24 +186,34 @@ type AggQueryPlan struct {
...
+// NewAggQueryPlan creates a new agg query plan
+func NewAggQueryPlan(qc *query.AQLQueryContext, topo topology.Topology, client dataCli.DataNodeQueryClient) (plan AggQueryPlan) {
+	var root common.MergeNode
+
+	shards := topo.Get().ShardSet().AllIDs()
+
+	// compiler already checked that only 1 measure exists, which is a expr.Call
+	measure := qc.Query.Measures[0].ExprParsed.(*expr.Call)
+	agg := common.CallNameToAggType[measure.Name]
+	if agg == common.Avg {
```
lucafuji: as said, probably we don't need this logic

broker/query_plan.go
shz117: yeah will add todo

query/expr/ast.go
shz117: yeah that's when we get `("foo", "bar")` in the query

broker/query_plan.go
shz117: because for each dimension, we need count from each data node to calculate the overall avg.
`totalAvt = (sum1 + sum2 +... + sumn) / (count1 + count2 + ... + countn)`

broker/query_plan.go
shz117: if a successful downstream call followed a failure, we clear error here and return the result.

broker/common/types.go
shz117: changed

broker/query_plan.go
shz117: addressed

broker/query_plan.go
lucafuji: should we consider break down avg and counts? because sometimes sum may cause overflow for some fields but we can still compute avg

shz117: hmm data nodes can calculate avg even if sum would overflow?

shz117: or even better: we can add a new agg function `rollingAvg`  at datanode level which returns avg and sum in 1 call.
let's keep it this way for now. will add todo for this one.

lucafuji: sounds good for me. Sometimes we can compute avg without the total sum , that's what we are doing right now. 
Because for some column sum may not make sense but avg does.

PR #201, opened 2019-06-10
Hash Reduction

lucafuji:
1. only Cuda part in this diff,  processor part coming later 
2. only support device mode, host concurrent hash map implementation will be done in future
3. concurrent hash map need c++14, we may consider to rewrite it ourselves.
4. hash vector and index vector no longer needed
5. include a device smart pointer implementation in this diff, may consider migrating other device memory allocation in future diffs


query/hash_reduction.cu
```diff
@@ -0,0 +1,395 @@
...
+        outputKeys.DimValues, dimWidthPrefixSum,
+        outputValue, globalWriteIndexDevice.get());
+
+    thrust::for_each_n(GET_EXECUTION_POLICY(cudaStream),
+                       thrust::counting_iterator<uint32_t>(0),
+                       length, extractorFunc);
+    // copy the reduced count back from GPU.
+    uint32_t
+    globalWriteIndexHost;
+    ares::asyncCopyDeviceToHost(
```
shz117: async + global seems fishy..

query/hash_reduction.cu
```diff
@@ -0,0 +1,395 @@
...
+  }
+
+  __host__ __device__
+  void operator()(const int i) {
+    key_type unusedKey = map->get_unused_key();
+    key_value_ptr_type mapValues = map->data();
+    size_t mapSize = map->capacity();
+    if (i < mapSize) {
+      key_type current_key = mapValues[i].first;
+      if (current_key != unusedKey) {
```
shz117: is it better to use `key_equal`?

lucafuji: no, key_equal here is compare the hash part (higher 32 bit) of the key, here we need to compare to whole key

shz117: i see.

query/hash_reduction.cu
lucafuji: yeah, you are right, we should use the synced copy here

query/hash_reduction.cu
```diff
@@ -0,0 +1,395 @@
...
+        sizeof(uint32_t), cudaStream);
+    return globalWriteIndexHost;
+  }
+
+  // execute reduces the dimension values with measures first and then extract
+  // aggregated result.
+  int execute(DimensionVector inputKeys, uint8_t *inputValues,
+              DimensionVector outputKeys, uint8_t *outputValues,
+              cudaStream_t cudaStream) {
+    auto equality = []
```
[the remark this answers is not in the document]
lucafuji: yes,  but since it's lambda so we cannot store it as a member function. 
And since the equality check logic is simple so I think it's ok to duplicate the logic. 

This is actually a problem of this concurrent hash map, on L297 of the concurrent_map.cuh
```
typename hash_value_type = typename Hasher::result_type
```
it requires result type of the Hasher, however since our hasher is a lambda, it does not have a hash_value_type bound with it.

Anyway I think to some days we probably need to rewrite the hash map ourselves.

query/hash_reduction.cu
lucafuji: or better to use cudaWaitForStream

PR #202, opened 2019-06-12
add datanode query client

shz117:


datanode/client/query_client.go
```diff
@@ -0,0 +1,87 @@
...
+	res, err = dc.client.Do(req)
+	if err != nil {
+		return
+	}
+	if res.StatusCode != http.StatusOK {
+		err = errors.New(fmt.Sprintf("got status code %d from datanode", res.StatusCode))
+		return
+	}
+	var respBytes []byte
+	respBytes, err = ReadAll(res.Body)
```
voyager-dw: shall we close the resp.body after read?

shz117: yeah good catch.

voyager-dw (approved):

PR #203, opened 2019-06-13
Purge check and change newmemstore/newtableshard to use options

voyager-dw:
Items included in this pr:
1. AcquireToken from bootstrapserver when do data purge
2. use options to replace multiple arguments in NewTableShard/NewMemStore


datanode/bootstrap/bootstrap_server.go
```diff
@@ -80,6 +83,7 @@ func NewPeerDataNodeServer(metaStore common.MetaStore, diskStore diskstore.DiskS
 		diskStore:          diskStore,
 		sessions:           make(map[int64]*sessionInfo),
 		tableShardSessions: make(map[tableShardPair][]int64),
+		tableShardTokens:   make(map[tableShardPair]int),
```
shz117: how is the token count different from len(tableShardSessions[key]) ?

datanode/bootstrap/bootstrap_server.go
```diff
@@ -100,15 +104,22 @@ func (p *PeerDataNodeServerImpl) AcquireToken(tableName string, shardID uint32)
...
 
 // AcquireToken release the token count, must call this when call AcquireToken success
 func (p *PeerDataNodeServerImpl) ReleaseToken(tableName string, shardID uint32) {
-	// nothing to do now
+	p.Lock()
+	defer p.Unlock()
+
+	key := tableShardPair{table: tableName, shardID: shardID}
+	if val, ok := p.tableShardTokens[key]; ok && val > 0 {
+		p.tableShardTokens[key] = p.tableShardTokens[key] - 1
```
shz117: should this also decrement on session expiration?

memstore/purge.go
```diff
@@ -17,10 +17,18 @@ package memstore
 import (
 	memCom "github.com/uber/aresdb/memstore/common"
 	"github.com/uber/aresdb/utils"
+	"github.com/pkg/errors"
 )
 
 // Purge purges out of retention data for table shard
 func (m *memStoreImpl) Purge(tableName string, shardID, batchIDStart, batchIDEnd int, reporter PurgeJobDetailReporter) error {
+	// check if there is peer bootstraping job running, skip this purge if we can not acquire the token
+	if m.options.bootstrapToken.AcquireToken(tableName, uint32(shardID)) {
```
shz117: do we need to AcquireToken for other processes as well (snapshot, archive) ?

voyager-dw: yes, both snapshot/archiving/backfill covered

diskstore/local_diskstore.go
```diff
@@ -211,7 +211,9 @@ func (l LocalDiskStore) OpenSnapshotVectorPartyFileForRead(table string, shard i
 	redoLogFile int64, offset uint32, batchID int, columnID int) (io.ReadCloser, error) {
 	snapshotFilePath := GetPathForTableSnapshotColumnFilePath(l.rootPath, table, shard, redoLogFile, offset, batchID, columnID)
 	f, err := os.OpenFile(snapshotFilePath, os.O_RDONLY, 0644)
-	if err != nil {
+	if os.IsNotExist(err) {
```
jshencode: why not just directly return os.ErrNotExist

memstore/purge.go
```diff
@@ -17,10 +17,18 @@ package memstore
...
+		defer m.options.bootstrapToken.ReleaseToken(tableName, uint32(shardID))
+	} else {
+		return errors.New("unable to acquire bootstrap token, retry later")
```
jshencode: return error will panic the server, we can just log here

diskstore/local_diskstore.go
voyager-dw: this is to follow the other convention, we not treat this is error

jshencode: this might be a different case, for user still check for reader == nil to get the os.ErrNonExist.  But for most metastore cases, user does not care about the os.ErrNonExist error

voyager-dw: then we need to change the behavior of both OpenVectorPartyFileForRead and OpenSnapshotFileForRead

jshencode: doesn't seem to change the behavior, before the change, all errors including nonexist error return to the user.
after the change, we just don't wrap the NonExistErr with stack error if user need to know this specific error

PR #204, opened 2019-06-14
broker working e2e for agg query

shz117:
- integrate with topology
 - implement broker schema mutator
 - update brokerd cmd and build
 - make broker namespace level


config/ares-broker.yaml
```diff
@@ -0,0 +1,33 @@
+port: 9474
```
voyager-dw: these config need to put into our internal broker code, remove asap

broker/broker_schema_mutator.go
```diff
@@ -50,34 +59,73 @@ func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err er
...
+		return
+	}
+	oldSchema.Columns[target].Config = config
+	b.tables[table] = memCom.NewTableSchema(&oldSchema)
 	return
 }
 
 func (b *BrokerSchemaMutator) DeleteColumn(table string, column string) (err error) {
+	oldSchema := b.tables[table].Schema
+	target := -1
```
voyager-dw: no lock control for all these schema operations?

broker/broker_schema_mutator.go
```diff
@@ -50,34 +59,73 @@ func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err er
...
+	if target == -1 {
+		err = errors.New(fmt.Sprintf("column %s not found", column))
+		return
+	}
+	oldSchema.Columns[target].Deleted = true
+	b.tables[table] = memCom.NewTableSchema(&oldSchema)
 	return
 }
 
 // ====  memstore/common.TableSchemaReader ====
```
voyager-dw: schemareader or schemamutator?

broker/broker_schema_mutator.go
shz117: not for now. since only 1 process (schemaFetchJob) will be updating them sequentially

broker/broker_schema_mutator.go
shz117: both, TableSchemaReader is defined and required in memstore.
TableSchemaMutator is defined in metastore and needed by the SchemaFetchJob

broker/broker_schema_mutator.go
```diff
@@ -31,12 +33,19 @@ type BrokerSchemaMutator struct {
...
+		tables: map[string]*memCom.TableSchema{},
+	}
 }
 
 // ====  metastore/common.TableSchemaMutator ====
-// TODO: implement. TableSchemaMutator should update b.tables
 func (b *BrokerSchemaMutator) ListTables() (tables []string, err error) {
+	tables = make([]string, len(b.tables))
+	i := 0
+	for _, table := range b.tables {
```
lucafuji:
```
for i, table := range b.tables {
		tables[i] = table.Schema.Name
	}
```

shz117: lol can't remember why i didn't do that..

broker/broker_schema_mutator.go
```diff
@@ -50,34 +59,73 @@ func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err er
...
+	b.tables[table] = memCom.NewTableSchema(&oldSchema)
 	return
 }
 
 func (b *BrokerSchemaMutator) DeleteColumn(table string, column string) (err error) {
+	oldSchema := b.tables[table].Schema
+	target := -1
+	for i, col := range oldSchema.Columns {
+		if col.Name == column {
+			target = i
```
lucafuji: better use sort.Search

broker/broker_schema_mutator.go
```diff
@@ -50,34 +59,73 @@ func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err er
...
+			break
+		}
+	}
+	if target == -1 {
+		err = errors.New(fmt.Sprintf("column %s not found", column))
+		return
+	}
+	oldSchema.Columns[target].Deleted = true
```
lucafuji: why not change the value in place? Seems like it's a pointer

lucafuji: you can avoid allocate a new object via this way

broker/broker_schema_mutator.go
shz117: oh I remembered. so b.tables is a map, i from the for iterator will be the key of string type.

broker/broker_schema_mutator.go
```diff
@@ -50,34 +59,73 @@ func (b *BrokerSchemaMutator) GetTable(name string) (table *common.Table, err er
...
 
 // ====  memstore/common.TableSchemaReader ====
-// TODO: implement. these are used by compiler
 func (b *BrokerSchemaMutator) GetSchema(table string) (tableSchema *memCom.TableSchema, err error) {
+	tableSchema = b.tables[table]
 	return
 }
 
 func (b *BrokerSchemaMutator) GetSchemas() (schemas map[string]*memCom.TableSchema) {
+	schemas = b.tables
```
lucafuji: for a single return value function like this.., I'm not sure the named return value does help reduce the code size or not..

broker/config/config.go
```diff
@@ -24,4 +25,6 @@ type BrokerConfig struct {
 
 	ControllerConfig *common.ControllerConfig `yaml:"controller,omitempty"`
 	HTTP             common.HTTPConfig        `yaml:"http"`
+	Etcd etcd.Configuration `yaml:"etcd"`
```
lucafuji: this file does not look like correct formatted

broker/broker_schema_mutator.go
shz117: because deleting a table will also cause `ValueTypeByColumn` to change, I didn't want to duplicate the logic in `NewTableSchema` here. better recreate by calling `NewTableSchema` since it's not a frequent call.

broker/broker_schema_mutator.go
shz117: sort.Search() requires sorted slice. we are deleting column by name, schema.Columns is not sorted by name.

broker/broker_schema_mutator.go
shz117: yeah I was just filling in the blanks, will change.

config/ares-broker.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+http:
+  max_connections: 300
+  read_time_out_in_seconds: 20
+  write_time_out_in_seconds: 300 # 5 minutes to write the result
+
+controller:
```
voyager-dw: is this for muttley, it also internal stuff. for oss, we just need use host:port. The http endpoint creation likely need some wrap

shz117: yes this is for muttley. but it's using http header which is generic enough for OSS. I can remove the uber specific header from config file though.

PR #206, opened 2019-06-14
Hash reduction processor hookup and host concurrent unordered map implementation

lucafuji:
1. the hash reduction may not work very well for non-atomic operations like avg. For those operations, we better use sort reduction
2. hash reduction may not work very well when the number of dim values is small.
3. in other cases, hash reduction beat sort reduction almost by 2x


cgoutils/memory.go
```diff
@@ -35,6 +35,10 @@ func IsPooledMemory() bool {
 	return (GetFlags() & C.POOLED_MEMORY_FLAG) != 0
 }
 
+func SupportHashReduction() bool {
```
shz117: feels like `SupportHashReduction` is not part of `memory.go`..

cgoutils/memory/malloc.c
```diff
@@ -19,7 +19,7 @@
 #include "../memory.h"
 
 DeviceMemoryFlags GetFlags(){
-  return 0x0;
+  return 0x4;
```
shz117: `return HASH_REDUCTION_SUPPORT` ?

query/aql_context.go
```diff
@@ -405,3 +407,13 @@ type AQLQueryContext struct {
 func (ctx *OOPKContext) IsHLL() bool {
 	return ctx.AggregateType == C.AGGR_HLL
 }
+
+func isAtomicAggType(aggType C.enum_AggregateFunction) bool {
+	return aggType == C.AGGR_SUM_SIGNED || aggType == C.AGGR_SUM_FLOAT;
```
shz117: how about `AGGR_SUM_UNSIGNED` for `count`?

query/aql_processor.go
```diff
@@ -765,14 +766,16 @@ func (bc *oopkBatchContext) prepareForDimAndMeasureEval(
 		})
 
 		// uint32_t for index value
-		bc.dimIndexVectorD = bc.reallocateResultBuffers(bc.dimIndexVectorD, 4, stream, nil)
+		if isHLL || !useHashReduction {
```
shz117: for hash reduction we don't need indexVector, right..?

cgoutils/memory.go
lucafuji: well, there is no better place to put those flags ..

query/aql_context.go
lucafuji: in cudf's map implementation, sum<uint32_t> is not  using atomic operations, therefore the performance will be bad. In some days we will rewrite the hash map and by that time,we can use hash reduction for count

shz117: i see, can we use uint64 for count if it's hash reduction..?

lucafuji: we can use int64, not worth changing the code now, we are gonna rewrite the hash table ourselves anyway

query/concurrent_unordered_map.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+#ifndef QUERY_CONCURRENT_UNORDERED_MAP_HPP_
+#define QUERY_CONCURRENT_UNORDERED_MAP_HPP_
+
+#if defined(SUPPORT_HASH_REDUCTION) && defined(RUN_ON_DEVICE)
+#include <hash/concurrent_unordered_map.cuh>
+#include <hash/hash_functions.cuh>
+#endif
+#include <thrust/pair.h>
+#if defined(SUPPORT_HASH_REDUCTION) && defined(RUN_ON_DEVICE)
```
shz117: merge with line 18?

lucafuji: lint requires those header files needs to be in order, cuh is considered as C header file while hpp are considered cpp header file

PR #208, opened 2019-06-17
peer source implementation

jshencode:
implementation of PeerSource interface
manages grpc connections to all peers in the cluster
use topology to add new peer and close old peers

major file changes:
datanode/peer_source.go

### DIFF
--- a/datanode/peer_source.go
+++ b/datanode/peer_source.go
@@ -1,0 +1,164 @@
+package datanode
+
+import (
+	"github.com/pkg/errors"
+	"github.com/uber/aresdb/cluster/topology"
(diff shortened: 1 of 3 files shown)


datanode/peer_source.go
```diff
@@ -0,0 +1,164 @@
...
+	close(ps.done)
+
+	ps.Lock()
+	defer ps.Unlock()
+	for _, peer := range ps.peers {
+		peer.Close()
+	}
+}
+
+func (ps *peerSource) updateTopoMap(topoMap topology.Map) {
```
voyager-dw: maybe we only need to get the shard peers

PR #211, opened 2019-06-17
non agg query execution

shz117:


datanode/client/query_client.go
```diff
@@ -46,7 +46,33 @@ type aqlRespBody struct {
...
+		err = errors.New(fmt.Sprintf("invalid response from datanode, resp: %s", bs))
+		return
+	}
+	result = respBody.Results[0]
+	utils.GetLogger().With("host", host, "query", query, "result", result).Debug("datanode query client Query succeeded")
+	return
+}
+
+func (dc *dataNodeQueryClientImpl) QueryRaw(ctx context.Context, host topology.Host, query queryCom.AQLQuery) (bs []byte, err error) {
+		bs, err = dc.queryRaw(ctx, host, query)
```
voyager-dw: lint?

broker/query_plan_agg.go
```diff
@@ -140,7 +140,7 @@ type ScanNode struct {
 	dataNodeClient dataCli.DataNodeQueryClient
 }
 
-func (sn *ScanNode) Execute(ctx context.Context) (result queryCom.AQLQueryResult, err error) {
+func (sn *BlockingScanNode) Execute(ctx context.Context) (result queryCom.AQLQueryResult, err error) {
 	trial := 0
 	for trial < rpcRetries {
```
voyager-dw: where is the rpcRetries definition?

datanode/client/query_client.go
shz117: already updated

broker/query_plan_agg.go
shz117: it's a const on the top of the file

PR #212, opened 2019-06-17
continue moving controller code to oss

jshencode:
1. MembershipMutator
2. DefaultKafkaConfig (can be override within internal repo)


controller/models/Instance.go
```diff
@@ -13,7 +13,83 @@
...
+	i = InActive
+	if string(data) == "active" {
+		i = Active
+	}
+	return nil
+}
+
+// Instance models an aresdb instance
+type Instance struct {
+	Name  string        `json:"name"`
```
voyager-dw: what is the difference between name and host here? or add some more comments

controller/models/Instance.go
```diff
@@ -13,7 +13,83 @@
...
+	Host  string        `json:"host"`
+	Port  int           `json:"port"`
+	State InstanceState `json:"state"`
+}
+
+// instanceInternal is a helper class to marshall and unmarshall Instance
+type instanceInternal struct {
```
voyager-dw: looks link no difference between instanceInternal and Instance

controller/models/Instance.go
```diff
@@ -13,7 +13,83 @@
...
+type InstanceView struct {
 	Address string `json:"address"`
 }
+
+// InstanceState define instance state
+type InstanceState int
+
+const (
+	// Active indicates an instance is healthy
+	Active InstanceState = iota
```
shz117: I defined it sort of arbitrarily. now that we have m3db integration, it may make sense to change the states to something in line with what topology defines.

controller/models/Instance.go
shz117: as commented, it was used to marshal Instance structs

controller/models/Instance.go
shz117: Name is an internal ID known by controller, host name is the physical host name. they are the same for now, but could be different in the future.

controller/models/Instance.go
jshencode: topology does not have state for host, actually i am not really seeing the state being actually used anywhere. maybe i will just keep the instance only to simplify since we will deprecate MembershipMutator eventually after migration.

shz117: no, it was not used. if the status will not be on hosts, we can remove it.

jshencode:
@shz117 @voyager-dw removed membership mutator and instance related code. will keep that part only for internal since we won't be needing that anyway after completely migrated to etcd

PR #218, opened 2019-06-27
broker refactor

shz117:
some refactoring to remove broker dependency on C libs
 - refactor compiler to remove dependency on `query` pkg
 - refactor `api` pkg to remove broker's dependency on `api` 
 - add shards in query, collapse shards to per host level


broker/query_compiler.go
```diff
@@ -0,0 +1,123 @@
...
+	if _, ok := c.AQLQuery.Measures[0].ExprParsed.(*expr.NumberLiteral); ok {
+		c.IsNonAggregationQuery = true
+		// in case user forgot to provide limit
+		if c.AQLQuery.Limit == 0 {
+			c.AQLQuery.Limit = nonAggregationQueryLimit
+		}
+		return
+	}
+
+	// TODO: handle limit properly
```
voyager-dw: wrong comment position?

shz117: not really.. not sure what's the best place to put this TODO, as it may require chages in multiple places. just put somewhere so I will remember.

PR #219, opened 2019-06-27
change uuid package to gofrs

jshencode:


jshencode:
> LGTM
> why the change btw?


according to david's diff 
```IMPORTANT: github.com/satori/go.uuid is no longer maintained and should not be used! https://github.com/gofrs/uuid is the recommended library for UUIDs and is backwards-compatible with satori, please consider updating!
```

PR #220, opened 2019-06-27
ShardOwner interface

jshencode:
use ShardOwner interface to remove hard coded shards

during query time, if query specified shards, use query specified shards for query
if query does not specify shards, use owned shards


broker/executor.go
```diff
@@ -56,7 +56,7 @@ func (qe *queryExecutorImpl) Execute(ctx context.Context, sqlQuery string, w htt
 	qc := &query.AQLQueryContext{
 		Query: aqlQuery,
 	}
-	qc.Compile(qe.tableSchemaReader)
+	qc.Compile(qe.tableSchemaReader, topology.NewTopologyShardOwner(qe.topo))
```
shz117: heads up: we won't need this after my PR merged

jshencode: ok. can remove that during the merge

datanode/datanode.go
```diff
@@ -469,7 +474,18 @@ func (d *dataNode) AssignShardSet(shardSet shard.ShardSet) {
 			}
 		}()
 	}
-	d.RUnlock()
+}
+
+// GetOwnedShards returns all shard ids the datanode owns
+func (d *dataNode) GetOwnedShards() []int {
```
voyager-dw: kind of duplicate as above (t *topologyShardOwner) GetOwnedShards, can we use the d.topo directkly?

jshencode: no topologyShardOwner gets all shards from the topology

this get the shards this host owns

PR #222, opened 2019-07-02
[broker] support aql 

shz117:


broker/common/types.go
```diff
@@ -29,10 +29,14 @@ type SchemaManager interface {
 	GetTableSchemaReader(namespace string) (memCom.TableSchemaReader, error)
 }
 
+// QueryContext defines query context
+type QueryContext interface {
```
jshencode: empty interface?

shz117: nah it's something I forgot to clean up.. thanks for the catch.

PR #223, opened 2019-07-02
Fix datanode

jshencode:
1. remove hard coded shard 0 pre-creation, instead use MkdirAll before all file writes
2. remove shard checking in MetaStore. metastore does not check for shard ownership anymore
3. fix AssignShardSet bug
4. add missing etcd client creation

### DIFF
--- a/memstore/memstore.go
+++ b/memstore/memstore.go
@@ -95,9 +95,9 @@ type memStoreImpl struct {
 
 	// reference to metaStore for registering watchers,
 	// fetch latest schema and store Shard versions.
-	metaStore            metaCom.MetaStore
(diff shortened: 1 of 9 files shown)


memstore/memstore.go
```diff
@@ -264,7 +264,7 @@ func (m *memStoreImpl) AddTableShard(table string, shardID int, needPeerCopy boo
 	defer m.Unlock()
 
 	schema := m.TableSchemas[table]
-	if schema != nil {
+	if schema == nil {
```
voyager-dw: error handling or output some error msg?

jshencode: this is not really an error

PR #224, opened 2019-07-02
fix schema fetch job

shz117:
- continue table processing after reporting failure
 - remove validation for new column with deletion error, as it's valid during migrations

### DIFF
--- a/metastore/schema_fetch.go
+++ b/metastore/schema_fetch.go
@@ -104,7 +104,7 @@ func (j *SchemaFetchJob) applySchemaChange(tables []common.Table) (err error) {
 			// found new table
 			err = j.schemaMutator.CreateTable(&table)
 			if err != nil {
-				reportError(err)
+				reportError(err, table.Name)
 				continue
 			}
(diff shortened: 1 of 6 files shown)


metastore/validator.go
```diff
@@ -85,10 +85,6 @@ func (v tableSchemaValidatorImpl) validateIndividualSchema(table *common.Table,
 	for columnID, column := range table.Columns {
 		if !column.Deleted {
 			nonDeletedColumnsCount++
-		} else {
-			if creation {
```
voyager-dw: we dont need to check?

metastore/schema_fetch.go
```diff
@@ -104,28 +104,33 @@ func (j *SchemaFetchJob) applySchemaChange(tables []common.Table) (err error) {
 			// found new table
 			err = j.schemaMutator.CreateTable(&table)
 			if err != nil {
-				return
+				reportError(err)
```
voyager-dw: can we add table name in the error report?

metastore/validator.go
shz117: it can be valid if the table is migrated from other clusters, or used for bootstraping, and have deleted columns.

PR #228, opened 2019-07-03
Hostport

shz117:


datanode/client/query_client.go
```diff
@@ -87,7 +88,12 @@ func (dc *dataNodeQueryClientImpl) QueryRaw(ctx context.Context, host topology.H
 }
 
 func (dc *dataNodeQueryClientImpl) queryRaw(ctx context.Context, host topology.Host, query queryCom.AQLQuery, hll bool) (bs []byte, err error) {
-	url := fmt.Sprintf("http://%s/query/aql?dataonly=1", host.Address())
+	// TODO: @shz remove this hack when we get proper address from host.Address()
+	address := host.Address()
+	if strings.HasPrefix(address, "http://") {
```
voyager-dw: better if you see any http:// https://, etc, then no need to prefix protocol
and how about sql query

shz117: updated.

sql query will use the same client

datanode/client/query_client.go
```diff
@@ -25,6 +25,7 @@ import (
 	"github.com/uber/aresdb/utils"
 	. "io/ioutil"
 	"net/http"
+	url2 "net/url"
```
jshencode: url2?

shz117: lol ide

PR #233, opened 2019-07-04
use random hash seed to balance sharding

voyager-dw:


subscriber/common/sink/sink.go
```diff
@@ -25,6 +25,11 @@ import (
 	"github.com/uber/aresdb/utils"
 )
 
+const (
+	// random hash seed, the default 0 is not well balanced
+	hashSeed = 115522763
```
shz117: but why is this seed different from 0? it's also a static number right?
to get something really random, maybe the seed should come from time, or /dev/random

voyager-dw: from the murmur code, using 0 as seed is very special, should use some random number instead. we can use this for testing first to see the balance result, or use other sharding algo if we see any issue.

PR #234, opened 2019-07-09
[broker] support limit and fix header issue

shz117:
- implement LIMIT support for non-agg query
 - column wildcard expansion in broker

### DIFF
--- a/broker/query_plan_non_agg.go
+++ b/broker/query_plan_non_agg.go
@@ -107,7 +107,10 @@ type NonAggQueryPlan struct {
 	resultChan chan streamingScanNoderesult
 	headers    []string
 	nodes      []*StreamingScanNode
-	limit      int
+	// number of rows needed
+	limit int
+	// number of rows flushed
+	flushed int
 }
 
 func (nqp *NonAggQueryPlan) Execute(ctx context.Context) (err error) {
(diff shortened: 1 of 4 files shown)


broker/query_plan_non_agg.go
```diff
@@ -148,12 +157,46 @@ func (nqp *NonAggQueryPlan) Execute(ctx context.Context) (err error) {
 			return
 		}
 		// write rows
-		nqp.w.Write(res.data)
-		if i != len(nqp.nodes)-1 {
+		if nqp.limit < 0 {
+			// when no limit, flush data directly
+			nqp.w.Write(res.data)
+		} else {
+			res.data = append([]byte("["), res.data[:]...)
```
voyager-dw: add some more test cases to simulate multiple nodes return and merge?

broker/query_plan_non_agg.go
```diff
@@ -140,6 +145,10 @@ func (nqp *NonAggQueryPlan) Execute(ctx context.Context) (err error) {
 	}
 
 	for i := 0; i < len(nqp.nodes); i++ {
+		if nqp.getRowsWanted() == 0 {
```
voyager-dw: better use notification or channel to find ready nqp to process return data immediately, instead of looping through nqps.

shz117: this is not really looping through nodes.. it's like a while loop, but we know before hand the max number of time we go into the loop will be number of child nodes.
the blocking channel read is at line 153

broker/query_plan_non_agg.go
shz117: it's in the query_plan_non_agg_test.go. simulated 3 nodes.

PR #235, opened 2019-07-09
refactor metrics

shz117:
to enable different components initial only subset of metrics

### DIFF
--- a/utils/metrics.go
+++ b/utils/metrics.go
@@ -21,11 +21,21 @@ import (
 	"sync"
 )
 
+// ReporterType is the type of reporter
+type ReporterType int
+
+// List of reporter types
+const (
+	ReporterTypeDataNode ReporterType = iota
+	ReporterTypeBroker
+)
+
 // MetricName is the type of the metric.
 type MetricName int
 
 // List of supported metric names.
 const (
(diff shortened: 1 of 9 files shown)


utils/metrics.go
```diff
@@ -892,6 +906,8 @@ var metricsDefs = map[MetricName]metricDefinition{
 	},
 }
 
+var brokerMetricsDefs = map[MetricName]metricDefinition{}
```
voyager-dw: empty?

utils/metrics.go
```diff
@@ -21,11 +21,21 @@ import (
 	"sync"
 )
 
+// ReporterType is the type of reporter
+type ReporterType int
+
+// List of reporter types
+const (
+	ReporterTypeDataNode ReporterType = iota
```
voyager-dw: shall we add subscriber and controller too?

shz117: subscriber and controller don't use this di yet.
we can easily add new types and sentinels when we migrate them to this

utils/metrics.go
shz117: yeah for now.
will add actual metrics when needed.

PR #236, opened 2019-07-09
add bootstrap details, logging and metrics

jshencode:


datanode/bootstrap/bootstrap_details.go
```diff
@@ -0,0 +1,84 @@
...
+
+type status int
+
+const (
+	notApplicable status = iota
+	copyNeeded
+	copyFinished
+)
+
+type bootstrapDetails struct {
```
voyager-dw: will update debug ui to output bootstrap detail?

jshencode: yeah. will add a new ui tab for bootstrap

PR #237, opened 2019-07-09
add metrics

shz117:


broker/handler.go
```diff
@@ -60,13 +80,30 @@ func (handler *QueryHandler) HandleSQL(w http.ResponseWriter, r *http.Request) {
...
-	err := apiCom.ReadRequest(r, &queryReqeust)
+	utils.GetRootReporter().GetCounter(utils.QueryReceivedBroker).Inc(1)
+
+	start := utils.Now()
+	var err error
+	defer func() {
+		duration := utils.Now().Sub(start)
+		utils.GetRootReporter().GetTimer(utils.QueryLatencyBroker).Record(duration)
+		if err != nil {
+			utils.GetRootReporter().GetCounter(utils.QueryFailedBroker).Inc(1)
```
voyager-dw: differentiate some metrics between aql/sql?

broker/query_plan_agg.go
```diff
@@ -109,8 +109,10 @@ func (mn *mergeNodeImpl) Execute(ctx context.Context) (result queryCom.AQLQueryR
 		}(i, c)
 	}
 
+	dataNodeWaitStart := utils.Now()
 	// TODO early merge before all results come back
 	wg.Wait()
+	utils.GetRootReporter().GetTimer(utils.TimeWaitedForDataNode).Record(utils.Now().Sub(dataNodeWaitStart))
```
voyager-dw: lint indent?

shz117: it has the same indent as the above line.. must be github ui rendering bug

broker/handler.go
shz117: I don't feel it's necessary to distinguish them.. main purpose for this is to trigger alerts on query failures no matter what caused.
if it's sql parsing error it will be logged.

voyager-dw: at least add some request count/error count for aql and sql to see how many query go through aql/sql in the future.

shz117: oh that makes sense. will add separate received counter.

PR #239, opened 2019-07-11
new shard on schema change

jshencode:
1. handle dimension table bootstrap
2. handle table addition event
3. clean up logging and bug fix

### DIFF
--- a/datanode/bootstrap.go
+++ b/datanode/bootstrap.go
@@ -30,17 +30,17 @@ import (
 	"time"
 
 	xerrors "github.com/m3db/m3/src/x/errors"
-	xsync "github.com/m3db/m3/src/x/sync"
 )
 
 // bootstrapManagerImpl is the implementation of the interface databaseBootstrapManager
 type bootstrapManagerImpl struct {
 	sync.RWMutex
 
-	datanode                    DataNode
-	peerSource                  client.PeerSource
 	opts                        Options
 	log                         common.Logger
+	origin                      string
+	peerSource                  client.PeerSource
+	bootstrapable               bootstrap.Bootstrapable
 	state                       bootstrap.BootstrapState
 	hasPending                  bool
 	bootstrapTableShards        map[string]bootstrap.BootstrapDetails
@@ -49,18 +49,19 @@ type bootstrapManagerImpl struct {
 }
 
 // NewBootstrapManager creates bootstrap manager
-func NewBootstrapManager(datanode DataNode, opts Options, topo topology.Topology) BootstrapManager {
+func NewBootstrapManager(origin string, bootstrappable bootstrap.Bootstrapable, opts Options, topo topology.Topology) BootstrapManager {
 	peerSource, err := NewPeerSource(topo)
 	if err != nil {
 		opts.InstrumentOptions().Logger().With("error", err.Error()).Fatal("failed to initalize peer source")
 	}
 
 	return &bootstrapManagerImpl{
-		datanode: datanode,
-		opts:     opts,
-		log:      opts.InstrumentOptions().Logger(),
-		peerSource: peerSource,
-		topo:     topo,
+		origin:        origin,
+		bootstrapable: bootstrappable,
+		opts:          opts,
+		log:           opts.InstrumentOptions().Logger(),
+		peerSource:    peerSource,
+		topo:          topo,
 	}
 }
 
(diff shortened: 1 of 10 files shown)


datanode/datanode.go
```diff
@@ -479,25 +553,49 @@ func (d *dataNode) AssignShardSet(shardSet shard.ShardSet) {
...
+			d.memStore.AddTableShard(table, int(shard.ID()), needPeerCopy)
 		}
 	}
+
+	// process dimension tables
+	// add dimension tables only if no previously existing shards owned by data node
+	// remove dimension tables only if no shard left for the node
+	if noExistingShards && !noShardLeft {
+		allShardsInitializing := initializingShards > 0 && len(incoming) == initializingShards
+		for _, table := range dimensionTables {
```
voyager-dw: what is the initializing logic here for dimension tables?

datanode/bootstrap.go
```diff
@@ -121,56 +123,20 @@ func (m *bootstrapManagerImpl) Bootstrap() error {
...
-				mutex.Unlock()
-				// unpin table shard after use
-				tableShard.Users.Done()
-			})
-		}
-	}
-	wg.Wait()
-	if !multiErr.Empty() {
-		return multiErr.FinalError()
+	err := m.bootstrapable.Bootstrap(m.peerSource, m.origin, m.topo, topoStateSnapshot, m.opts.BootstrapOptions())
```
shz117: is BootstrapManager and Bootstrap 1:1 ?
in that case BootstrapManager seems dummy and can be removed..

datanode/datanode.go
```diff
@@ -257,20 +258,43 @@ func (d *dataNode) startDebugServer() {
...
+	shardOwnershipEvents, done, err := d.metaStore.WatchShardOwnershipEvents()
+	if err != nil {
+		utils.GetLogger().With("error", err.Error()).Fatal("failed to watch schema addition")
+	}
+
+	for {
+		select {
+		case <-d.close:
+			return
+		case event, ok := <-shardOwnershipEvents:
```
shz117: curious what the second return value from channel read is

datanode/datanode.go
```diff
@@ -303,39 +327,52 @@ func (d *dataNode) startAnalyzingShardAvailability() {
...
+				factTables = append(factTables, table)
+			}
+		}
+		d.memStore.RUnlock()
+
 		// check all whether all tables within initialing shards are bootstrapped
 		for shardID := range initializing {
 			numTablesBootstrapped := 0
-			for _, table := range tables {
+			for _, table := range factTables {
```
shz117: who marks dimension table availability?

datanode/datanode.go
```diff
@@ -432,40 +456,90 @@ func (d *dataNode) advertise() {
...
-		d.opts.InstrumentOptions().Logger().Info("start advertising datanode to cluster",
-			zap.String("id", d.hostID))
+		d.logger.Info("start advertising datanode to cluster")
 	}
 }
 
-// GetTableShard returns the table shard from the datanode
-func (d *dataNode) GetTableShard(table string, shardID uint32) (*memstore.TableShard, error) {
-	return d.memStore.GetTableShard(table, int(shardID))
+func (d *dataNode) addTable(table string) {
```
voyager-dw: any conflict or change with the WatchShardOwnershipEvents in recovery?

datanode/bootstrap.go
jshencode: bootstrapManager still acts as a scheduler

datanode/datanode.go
jshencode: when channel is closed, ok will be false

datanode/datanode.go
jshencode: under current m3 framework we don't have availability tracking for dimension table (non sharded) out of box. we will need to add dimension table availability tracking ourselves in etcd

datanode/datanode.go
jshencode: in distributed mode, we do not call InitShards as we do in single instance mode under MemStore (in recovery.go)

datanode/datanode.go
jshencode: dimension tables will be added when first shard is assigned to the data node
and removed when the last shard is removed from the data node

datanode/datanode.go
shz117: I see. maybe add a todo somewhere

PR #242, opened 2019-07-12
hll EncodeBinary

shz117:


query/common/hll.go
```diff
@@ -688,11 +688,35 @@ func (hll *HLL) Encode() []byte {
...
+	}
+	data := make([]byte, (recordValueBytes+paddingBytes)*len(hll.SparseData))
 	for i, register := range hll.SparseData {
-		data[i*3] = byte(register.Index & 0xff)
-		data[i*3+1] = byte(register.Index >> 8)
-		data[i*3+2] = register.Rho
+		if padding {
+			data[i*recordValueBytes+paddingBytes] = byte(0x00)
+		}
+		data[i*(recordValueBytes+paddingBytes)+paddingBytes] = byte(register.Index & 0xff)
```
jshencode: can you make sure the order of bytes is the same with the in gpu format?

shz117: thanks for the catch.
updated.

query/common/hll.go
```diff
@@ -688,11 +688,36 @@ func (hll *HLL) Encode() []byte {
...
+	if padding {
+		recordValueBytes = 4
+	}
+	data := make([]byte, (recordValueBytes)*len(hll.SparseData))
 	for i, register := range hll.SparseData {
-		data[i*3] = byte(register.Index & 0xff)
-		data[i*3+1] = byte(register.Index >> 8)
-		data[i*3+2] = register.Rho
+		if padding {
+			data[i*recordValueBytes] = byte(0)
```
jshencode: actually better to do *(*uint32)(unsafe.Pointer(&data[i*recordValueBytes])) = uint32(register.Rho << 16 | register.Index), also could you add a test using readHLL to read the byte value back to HLL?

shz117: register.Rho << 16 won't really work.. 
even if it works, `|` between byte and uint16 wont' work..
it will need a lot of padding 0s and cast to uint32 for both `Rho` and `Index`, won't be cleaning than current way

shz117: I can add some comment to make it readable

shz117: nvm, looks like I just need to cast the 2 parts separately..

query/common/hll.go
```diff
@@ -814,3 +845,271 @@ var hllBiases = []float64{
...
+					dimVectors[level] = append(dimVectors[level], (*[4]byte)(byteVal)[:]...)
+				case 2:
+					dimVectors[level] = append(dimVectors[level], (*[2]byte)(byteVal)[:]...)
+				case 1:
+					dimVectors[level] = append(dimVectors[level], (*[1]byte)(byteVal)[:]...)
+				}
+			} else {
+				switch dimValueBytes {
+				case 8:
+					dimVectors[level] = append(dimVectors[level], make([]byte, 8)...)
```
jshencode: this allocation seems unnecessary. we can just append number of bytes directly and no need for switch case

shz117: true, this can be done in runtime without casting like the one above..

query/common/hll.go
```diff
@@ -688,11 +697,33 @@ func (hll *HLL) Encode() []byte {
...
+// aligns to 4 bytes for sparse hll
+// used to build response for application/hll queries from HLL struct
+func (hll *HLL) EncodeBinary() []byte {
+	if len(hll.DenseData) != 0 {
+		return hll.DenseData
+	}
+	return hll.encodeSparse(true)
+}
+
+func (hll *HLL) encodeSparse(padding bool) []byte {
```
jshencode: let's maybe add some comment about the two ways of encoding format
1. padding (1 byte) | rho (1 byte) | index (2 byte)   endianness based on the machine itself
2.                                rho (1 byte) | index (2 byte)  small endian

query/common/hll.go
```diff
@@ -814,3 +845,262 @@ var hllBiases = []float64{
...
+	}
+	for _, idx := range dimensionVectorIndex {
+		dimVector = append(dimVector, validityVectors[idx]...)
+	}
+
+	return
+}
+
+// helper function that traverses HLL query result
+func traverseRecursive(level int, curr interface{}, dimDataTypes []memCom.DataType, enumDicts map[int]map[string]int, hllVector, countVector *[]byte, dimVectors, validityVectors [][]byte) (err error) {
```
jshencode: level -> dimIndex ?

query/common/hll.go
```diff
@@ -814,3 +845,262 @@ var hllBiases = []float64{
...
+			var (
+				dataVal memCom.DataValue
+				isValid uint8 = 1
+			)
+			if NullString == k {
+				isValid = 0
+			} else {
+				if enumDict, ok := enumDicts[level]; ok {
+					enumVal := enumDict[k]
+					if dimValueBytes == 1 {
```
jshencode: we have a ConvertValueForType call in memCom does similar thing

query/common/hll.go
```diff
@@ -814,3 +845,262 @@ var hllBiases = []float64{
...
+						return
+					}
+				}
+			}
+
+			if dataVal.Valid {
+				byteVal := dataVal.OtherVal
+				switch dimValueBytes {
+				case 8:
+					dimVectors[level] = append(dimVectors[level], (*[8]byte)(byteVal)[:]...)
```
jshencode: actually we need to append the dim value multiple times depending on the next level result size

shz117: yep, fixed..

query/common/hll.go
shz117: hmm doesn't ConvertValueForType returns `interface{}` which I need to case again based on dimValueBytes?

query/common/hll.go
```diff
@@ -814,3 +848,281 @@ var hllBiases = []float64{
...
+		dimVector = append(dimVector, validityVectors[idx]...)
+	}
+
+	return
+}
+
+// helper function that traverses HLL query result in post order
+// returns: size is the total num dim rows of the subtree rooted by current node
+func traverseRecursive(dimIdx int, curr interface{}, dimDataTypes []memCom.DataType, enumDicts map[int]map[string]int, hllVector, countVector *[]byte, dimVectors, validityVectors [][]byte) (size int, err error) {
+	switch curr.(type) {
```
jshencode: we can use switch v := curr.(type)

PR #243, opened 2019-07-12
unify cluster config

jshencode:
need to modify internal config file


cmd/aresd/cmd/cmd.go
```diff
@@ -135,19 +135,19 @@ func start(cfg common.AresServerConfig, logger common.Logger, queryLogger common
 
 	// fetch schema from controller and start periodical job
 	if cfg.Cluster.Enable {
-		if cfg.Cluster.ClusterName == "" {
+		if cfg.Cluster.Namespace == "" {
```
voyager-dw: likely need to make sure not break the current production config

jshencode: will change config during merge and deploy

PR #248, opened 2019-07-16
bootstrap non-initializing shard first

jshencode:
1. bootstrap non initializing (available and leaving) shards first
2. on server start, only wait for non initializing shards readiness

### DIFF
--- a/datanode/datanode.go
+++ b/datanode/datanode.go
@@ -77,7 +77,7 @@ type dataNode struct {
 	mapWatch topology.MapWatch
 	close    chan struct{}
 
-	readyCh  chan struct{}
+	readyCh chan struct{}
 }
 
 type datanodeHandlers struct {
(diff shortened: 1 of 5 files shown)


datanode/datanode.go
```diff
@@ -302,83 +306,140 @@ func (d *dataNode) startActiveTopologyWatch() {
 	}
 }
 
-func (d *dataNode) startAnalyzingShardAvailability() {
-	ticker := time.NewTicker(5 * time.Second)
+func (d *dataNode) checkShardReadiness(tables []string, shards []uint32) (numReadyShards int) {
+	// check all whether all tables within initialing shards are bootstrapped
```
voyager-dw: update comment?

datanode/datanode.go
```diff
@@ -302,83 +306,140 @@ func (d *dataNode) startActiveTopologyWatch() {
...
+				continue
+			}
+			if tableShard.IsBootstrapped() {
+				numTablesBootstrapped++
+			}
+			tableShard.Users.Done()
+		}
+
+		if numTablesBootstrapped == len(tables) {
+			shards[i], shards[numReadyShards] = shards[numReadyShards], shards[i]
```
voyager-dw: what purpose of this line?

datanode/datanode.go
```diff
@@ -396,6 +457,9 @@ func (d *dataNode) ID() string {
 }
 
 func (d *dataNode) Serve() {
+	// wait for server is ready to serve
```
voyager-dw: add back the timer to see how much time take to recover?

datanode/datanode.go
jshencode: swap two element

this is part of the partition algorithm, which partitions the list into two parts, ready shards and non ready shards

datanode/datanode.go
jshencode: it is already in the Serve function

PR #251, opened 2019-07-22
Array support 2

voyager-dw:
add full functions support for list archive vp and list live vp(getvalue, setvalue, read, write, etc.)


memstore/common/data_type.go
```diff
@@ -583,6 +583,9 @@ func ArrayValueFromArray(value []interface{}, dataType DataType) (interface{}, e
 // we can also support comma delimited string value as long as the item not contain comma
 // "11,12,13"
 func ArrayValueFromString(value string, dataType DataType) (interface{}, error) {
+	if len(value) == 0 || value == "null" {
```
shz117: does null has to be all lower case?
maybe create a constant for null value string

memstore/common/vector_party.go
```diff
@@ -201,17 +201,8 @@ type ArchiveVectorParty interface {
 	SliceIndex(lowerBoundRow, upperBoundRow int) (startIndex, endIndex int)
 }
 
-// ListDataValueReader defines how to read ith element in jth row from a data source.
-// Out of boundary access will panic.
-type ListDataValueReader interface {
-	ReadElementValue(row int, i int) unsafe.Pointer
-	ReadElementBool(row int, i int) bool
-	ReadElementValidity(row int, i int) bool
```
shz117: we don't have element validity any more?

memstore/common/data_type.go
jshencode: does null represent a null list or a list with one element which is null?

voyager-dw: for simplicity, will treat null list and empty list both as null for now

voyager-dw: null is used in json parser, and we have other place use it too. yeah, create a constant will be better.

memstore/common/vector_party.go
voyager-dw: element is not needed in go lang, there is a ArrayValueReader which is only for testing purpose

memstore/list/archive_vector_party.go
```diff
@@ -0,0 +1,367 @@
...
+	bytesWritten    int64
+	totalValueBytes int64
+	// Used in archive batches to allow requesters to wait until the vector party
+	// is fully loaded from disk.
+	Loader sync.WaitGroup
+	// For archive store only. Number of users currently using this vector party.
+	// This field is protected by the batch lock.
+	pins int
+	// For archive store only. The condition for pins to drop down to 0.
+	allUsersDone *sync.Cond
```
jshencode: isn't allUsersDone already in Pinnable?

voyager-dw: good catch, will remove them

memstore/list/archive_vector_party.go
```diff
@@ -0,0 +1,367 @@
...
+	newBytes := common.CalculateListElementBytes(vp.dataType, newLen)
+	// Set offset.
+	vp.offsets.SetValue(2*row, unsafe.Pointer(&vp.bytesWritten))
+	// Set length.
+	vp.offsets.SetValue(2*row+1, unsafe.Pointer(&newLen))
+
+	baseAddr := uintptr(vp.values.Buffer()) + uintptr(vp.bytesWritten)
+	to := cgoutils.MakeSliceFromCPtr(baseAddr, newBytes)
+	from := cgoutils.MakeSliceFromCPtr(uintptr(val), newBytes)
+	for i := 0; i < newBytes; i++ {
```
jshencode: i think we can use copy() for this operation

memstore/list/archive_vector_party.go
```diff
@@ -0,0 +1,367 @@
...
+	newLen := int(*(*uint32)(val))
+	newBytes := common.CalculateListElementBytes(vp.dataType, newLen)
+	// Set offset.
+	vp.offsets.SetValue(2*row, unsafe.Pointer(&vp.bytesWritten))
+	// Set length.
+	vp.offsets.SetValue(2*row+1, unsafe.Pointer(&newLen))
+
+	baseAddr := uintptr(vp.values.Buffer()) + uintptr(vp.bytesWritten)
+	to := cgoutils.MakeSliceFromCPtr(baseAddr, newBytes)
+	from := cgoutils.MakeSliceFromCPtr(uintptr(val), newBytes)
```
jshencode: caution in using MakeSliceFromCPtr, the pointer has to be C allocated only. or the underlying address might be changed by golang garbage collector

use our own memcopy can achieve the same

memstore/list/archive_vector_party.go
```diff
@@ -0,0 +1,367 @@
...
+	vp.totalValueBytes = int64(bytes)
+	// Read value vector.
+	vp.values = memstore.NewVector(common.Uint8, int(vp.totalValueBytes))
+	// Here we directly read from reader into the c allocated bytes.
+	if err = dataReader.Read(cgoutils.MakeSliceFromCPtr(uintptr(vp.values.Buffer()), vp.values.Bytes)); err != nil {
+		return err
+	}
+
+	if s != nil {
+		s.ReportVectorPartyMemoryUsage(int64(vp.length*4*2) + vp.totalValueBytes)
```
jshencode: add comment on the calculation of memory usage

memstore/list/live_vector_party.go
```diff
@@ -0,0 +1,394 @@
...
+		} else {
+			dataWriter.WriteUint32(uint32(valueBytes))
+		}
+		dataWriter.WriteUint32(length)
+
+		valueBytes += common.CalculateListElementBytes(vp.dataType, int(length))
+	}
+	// to compatible with archive vp, align to 64 bytes alignment
+	if vp.offsets.Bytes > 8*vp.length {
+		dataWriter.SkipBytes(vp.offsets.Bytes - 8*vp.length)
```
jshencode: there is a call called WritePadding we can use

memstore/list/live_vector_party.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+
+	// value bytes, to compatible with archive vp, align to 64 bytes alignment
+	totalValueBytes := (valueBytes*8 + 511) / 512 * 64
```
jshencode: same use WritePadding for padding

memstore/list/live_vector_party.go
```diff
@@ -0,0 +1,394 @@
...
+	vp.offsets.SetValue(2*row, unsafe.Pointer(&buf[0]))
+	// Set length.
+	vp.offsets.SetValue(2*row+1, unsafe.Pointer(&newLen))
+	// Set footer offset.
+	vp.caps.SetValue(row, unsafe.Pointer(&buf[1]))
+
+	if valid {
+		baseAddr := vp.memoryPool.Interpret(buf[0])
+
+		to := cgoutils.MakeSliceFromCPtr(baseAddr, newBytes)
```
jshencode: same as archive vector party use memcpy directly

voyager-dw: changed

memstore/list/live_vector_party.go
voyager-dw: good one, updated

memstore/common/data_type.go
voyager-dw: will add special null vector in another diff

PR #257, opened 2019-07-26
BorrowConnection from multiple peers and pick the first healthy peer

jshencode:
…tion


datanode/client/types.go
```diff
@@ -18,10 +18,14 @@ import (
...
 	"github.com/uber/aresdb/datanode/generated/proto/rpc"
 	queryCom "github.com/uber/aresdb/query/common"
+	"google.golang.org/grpc"
 )
 
 // WithConnectionFn defines function with PeerDataNodeClient
-type WithConnectionFn func(rpc.PeerDataNodeClient)
+type WithConnectionFn func(peerID string, client rpc.PeerDataNodeClient)
+
+// PeerConnDialer defines the dial function for PeerDataDodeClient
```
voyager-dw: PeerDataDodeClient -> PeerDataNodeClient

voyager-dw: same here

PR #258, opened 2019-07-26
add health checking topology

shz117:
- add HealthTrackingDynamicTopology
 - move topology interfaces to common

next: mark host healhiness in broker

### DIFF
--- a/cluster/topology/healthtracking_dynamic.go
+++ b/cluster/topology/healthtracking_dynamic.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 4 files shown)


cluster/topology/healthtracking_dynamic.go
```diff
@@ -0,0 +1,133 @@
...
+
+	topo := &healthTrackingDynamicTopoImpl{
+		dynamicTopology:  dynamicTopo,
+		hostsHealthiness: make(map[Host]*healthiness),
+	}
+
+	return topo, nil
+}
+
+func (ht *healthTrackingDynamicTopoImpl) Get() Map {
```
voyager-dw: what is the difference of this class/function with existing StaticTopology or function automatically filter out unhealth nodes?

shz117: existing one depends on etcd as the source of truth for datanode healthiness. it depends on several things that will NOT affect whether queries can be served:
 1. network connection between datanode and etcd
 2. network connection between broker and etcd
Using HealthTrackingTopology, broker can track the healthiness itself.

E.g. consider a scenario where: 
  - all datanodes are alive
  - connections between datanode and etcd are lost
  - connections between broker and etcd alive
  - connections between broker and datanode are alive
If we use DynamicTopology with `includeUnhealthy = false`, we would not be able to serve any query.
with HealthTrackingTopology we can still serve queries.

cluster/topology/healthtracking_dynamic.go
```diff
@@ -0,0 +1,133 @@
...
+	if !found {
+		return utils.StackError(nil, "failed to change host health state, host not found. host: %s, healthiness %t", host, healthy)
+	}
+	h.healthy = healthy
+	h.lastUpdateTimestamp = utils.Now()
+	return nil
+}
+
+// dummy implementation, don't use
+func (ht *healthTrackingDynamicTopoImpl) Watch() (MapWatch, error) {
```
jshencode: TODO for this implementation? seems useful

PR #265, opened 2019-08-01
implement subscriber mutator and membership mutator for etcd version …

jshencode:
1. implement subscriber mutator and membership mutator for etcd version
2. remove watcher for subscriber heartbeat and use only pull mode and hash for assignment calculation to avoid eager recalculation of assignment during subscriber restart


controller/mutators/etcd/membership_mutator.go
```diff
@@ -0,0 +1,127 @@
...
+		etcdClient: etcdClient,
+	}
+}
+
+type membershipMutatorImpl struct {
+	etcdClient *kvstore.EtcdClient
+}
+
+func (mm membershipMutatorImpl) Join(namespace string, instance models.Instance) error {
+	return errors.New("join not supported in etcd version of membership mutator")
```
shz117: we depend on datanode heart beat instead?

jshencode: yeah. easiest route i think

jshencode: it is not through controller, each datanode heartbeat using m3db's heartbeat service

voyager-dw: let's have some talk to make it clear

PR #266, opened 2019-08-01
Scfg

lxning:
1. set string column length
2. fetch table schema and enum by demand


client/connector.go
```diff
@@ -287,18 +285,33 @@ func (u *UpsertBatchBuilderImpl) prepareEnumCases(tableName, columnName string,
...
-			if caseInsensitive {
-				enumCase = strings.ToLower(enumCase)
+			if len(enumCase) > defaultStringEnumLength {
+				abandonRow = true
+				u.logger.With(
+					"name", "prepareEnumCases",
+					"error", "Enum string value is too long",
+					"table", tableName,
+					"columnID", columnID,
+					"value", value).Debug("Enum string value is too long")
```
jshencode: we also might want some metrics for this case for monitoring

lxning: added

client/schema.go
```diff
@@ -94,27 +94,20 @@ func NewHttpSchemaFetcher(httpClient http.Client, address string, scope tally.Sc
 }
 
 // Start starts the CachedSchemaHandler, if interval > 0, will start periodical refresh
-func (cf *CachedSchemaHandler) Start(interval int) error {
-	err := cf.FetchAllSchema()
-	if err != nil {
-		return err
-	}
+func (cf *CachedSchemaHandler) Start(interval int) {
+	cf.FetchSchemas()
```
shz117: isn't this always a no-op?
cf.schemas will start with empty map, only Insert() calls will start fetching schema lazily.

lxning: Yes, this is my intention to make ares-subscriber initialization fast. Fetch schema and enum is by demand.

shz117: yeah then we can remove this call from the code.

lxning: I keep this b/c in case the other features will set tables in cachedSchemaHandler during initialization in the future. Then Start() will not delay the fetch in that case. And also in the current use case, the cost of this call can be ignored.

PR #275, opened 2019-08-07
support ad-hoc polygons in query

shz117:
- support ad-hoc polygon in a new filter `in_polygon(<point_column>, <polygon in WKT format>)`
 - allow exporting polygon as virtual dimension "_geoshape"


query/aql_compiler.go
```diff
@@ -967,6 +975,42 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
...
 
 			e.ExprType = expr.Boolean
+		case expr.PointInPolygonCallName:
+			if len(e.Args) != 2 {
+				qc.Error = utils.StackError(
+					nil, "expect 2 arguments for %s, but got %s", e.Name, e.String())
+				break
+			}
+			lhsRef, isVarRef := e.Args[0].(*expr.VarRef)
+			if !isVarRef || (lhsRef.DataType != memCom.GeoShape && lhsRef.DataType != memCom.GeoPoint) {
```
jshencode: lhs can be GeoShape?

query/aql_compiler.go
```diff
@@ -967,6 +975,42 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
 			}
 
 			e.ExprType = expr.Boolean
+		case expr.PointInPolygonCallName:
```
jshencode: can we still use geo_intersect call for this case

shz117: it's going to be tricking to reuse the name. as in qc.rewrite, we validate arg types base on call name, but not where the expr came from.

query/aql_compiler.go
shz117: nope. good catch.

shz117:
close for now, might need major refactor

PR #277, opened 2019-08-12
Suns

lxning:
1. Update etcdConfig to support uns 
2. Update etcdConfig loading
3. Refactor etcd heartbeat registration
4. Only load aresdb instances from jobAssignment if sinkmode is aresDB
5. Update GetDataValue if type is not string.
6. Set sarama publisher manual partition for fact table

### DIFF
--- a/subscriber/common/job/controller.go
+++ b/subscriber/common/job/controller.go
@@ -174,11 +174,12 @@ func connectEtcdServices(params Params) (services.Services, error) {
 		SetMetricsScope(params.ServiceConfig.Scope)
 
 	// etcd key format: prefix/${env}/namespace/service/instanceId
-	params.ServiceConfig.EtcdConfig.EtcdConfig.Env = fmt.Sprintf("%s/%s",
+	etcdConfig := &params.ServiceConfig.EtcdConfig.EtcdConfig
+	etcdConfig.Env = fmt.Sprintf("%s/%s",
 		params.ServiceConfig.EtcdConfig.EtcdConfig.Env, config.ActiveJobNameSpace)
 
 	// create a config service client to access to the etcd cluster services.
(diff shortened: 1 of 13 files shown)


subscriber/common/job/controller.go
```diff
@@ -517,6 +507,34 @@ func (c *Controller) startDriver(
...
+
+// RestartEtcdHBService registers heartbeat again if etcd cluster changes are detected
+func (c *Controller) RestartEtcdHBService(params Params) {
+	for {
+		select {
+		case <-config.EtcdCfgEvent:
+			// TODO: unadevertises old heartbeat and closes etcd client should be added once M3 provides
+			c.serviceConfig.Logger.Info("RestartEtcdHBService")
+			c.startEtcdHBService(params)
+		default:
```
jshencode: having default here turns this into a busy loop as opposed to go routine put into sleep until etcd event actually arrive

subscriber/common/sink/sink.go
```diff
@@ -124,16 +123,136 @@ func getPrimaryKeyBytes(row client.Row, destination Destination, jobConfig *rule
...
+		return memCom.ValueFromString(dataStr, dataType)
+	}
 
-	return dataVal, err
+	val := memCom.DataValue{
+		DataType: dataType,
+	}
+
+	var ok bool
+	switch dataType {
```
jshencode: feel like this section of code is useful for converting interface{} to DataValue, which can be moved into memstore/common/data_value.go so that we can consolidate later

lxning: done

subscriber/common/job/controller.go
lxning: done

lxning:
> could you also add the ManualPartitioner setting into the producer setting in this diff

done

PR #284, opened 2019-08-22
Conn

lxning:
1. add Close in client/connector
2. set controller default FailureHandlerMaxRetryMinutes as 15min instead of 365days


voyager-dw:
is this the cause of message lost?

lxning:
> is this the cause of message lost?

There was no message lost b/c the offset was not commited. Subscriber kept retrying since the failure handler retry window is 365days.

PR #289, opened 2019-08-27
Array support 4

voyager-dw:
binder/functor and go part change for array support,
will use separate diff to update null support and UUID support
more tests to add


memstore/list/archive_vector_party.go
```diff
@@ -379,6 +380,50 @@ func (vp *ArchiveVectorParty) SliceIndex(lowerBoundRow, upperBoundRow int) (
...
+	offsetStart := unsafe.Pointer(uintptr(vp.offsets.Buffer()) + uintptr(startIndex*8))
+
+	baseAddr := uintptr(vp.values.Buffer())
+	var valueStart int
+	valueBytes := vp.values.Bytes
+
+	for i := startIndex; i < (startIndex+length) && i < vp.length; i++ {
+		// find first entry which has non-zero length array value, which will have valid offset
+		// if not found, then will start from baseAddr
+		offset, length := vp.GetOffsetLength(i)
```
jshencode: maybe consider a different variable name for array length vs. vp length.

query/aql_processor_test.go
```diff
@@ -2342,4 +2370,18 @@ var _ = ginkgo.Describe("aql_processor", func() {
 		qc.initializeNonAggResponse()
 		Ω(w.Body.String()).Should(Equal(``))
 	})
+
+	ginkgo.It("Copy array to device should work", func() {
```
shz117: this test doens't really copy anything to device, right..?
seems it's actually testing `readDeviceVPSlice`  which is a test util function

query/aql_processor.go
```diff
@@ -1321,6 +1321,24 @@ func (qc *AQLQueryContext) runBatchExecutor(e BatchExecutor, isLastBatch bool) {
 
 // copyHostToDevice copy vector party slice to device vector party slice
 func copyHostToDevice(vps memCom.HostVectorPartySlice, deviceVPSlice deviceVectorPartySlice, stream unsafe.Pointer, device int) (bytesCopied, numTransfers int) {
+	if memCom.IsArrayType(vps.ValueType) {
+		// for array data type
+		// copy offset-length
+		cgoutils.AsyncCopyHostToDevice(
+			deviceVPSlice.basePtr.getPointer(), vps.Offsets, vps.Length*8,
```
jshencode: we can add a device pointer (offsets) instead of directly use basePtr for readability although it is the same as basePtr.

query/aql_processor.go
```diff
@@ -1346,6 +1364,22 @@ func copyHostToDevice(vps memCom.HostVectorPartySlice, deviceVPSlice deviceVecto
 }
 
 func hostToDeviceColumn(hostColumn memCom.HostVectorPartySlice, device int) deviceVectorPartySlice {
+	if memCom.IsArrayType(hostColumn.ValueType) {
+		deviceColumn := deviceVectorPartySlice{
+			length:       hostColumn.Length,
+			valueType:    hostColumn.ValueType,
+			defaultValue: hostColumn.DefaultValue,
+		}
+		totalColumnBytes := hostColumn.ValueBytes + hostColumn.Length*8
```
jshencode: comment with offset vector bytes for the length * 8 part or just add a variable called OffsetBytes

query/binder.hpp
```diff
@@ -357,6 +357,41 @@ class InputVectorBinderBase {
             defaultValueStruct.Value.Int64Val, int64_t)
         default: break;
       }
+    } else if (input.Type == ArrayVectorPartyInput) {
+      // Array type can only bind to first argument for BinaryTransformer
+      ArrayVectorPartySlice inputVP = input.Vector.ArrayVP;
+      uint8_t *basePtr = inputVP.OffsetLength;
```
jshencode: nit: may be rename OffsetLength to OffsetLengthVector?

query/functor.hpp
```diff
@@ -447,6 +447,150 @@ struct GetHLLValueFunctor {
...
+  __host__ __device__
+  result_type operator()(argument_type_1 arrVal, argument_type_2 indexT) const {
+    auto valid = thrust::get<1>(arrVal);
+    if (!valid) {
+      O v;
+      return thrust::make_tuple<O, bool>(v, false);
+    }
+    uint32_t *lenP = reinterpret_cast<uint32_t *>(thrust::get<0>(arrVal));
+    int index = thrust::get<0>(indexT);
+    if (lenP == NULL || (index >= 0 && *lenP <= index) ||
```
jshencode: use nullptr instead of NULL

query/functor.hpp
```diff
@@ -447,6 +447,150 @@ struct GetHLLValueFunctor {
...
+       (index < 0 && *lenP < -index)) {
+      O v;
+      return thrust::make_tuple<O, bool>(v, false);
+    }
+    auto valP = reinterpret_cast<I>(
+        reinterpret_cast<uint8_t*>(thrust::get<0>(arrVal)) + 4);
+    int len = static_cast<int>(*lenP);
+    if (index < 0) {
```
jshencode: are we supporting negative index as well? just curious

query/functor.hpp
```diff
@@ -447,6 +447,150 @@ struct GetHLLValueFunctor {
...
+    if (len == 0 || index >= len || index < 0) {
+        O v;
+        return thrust::make_tuple<O, bool>(v, false);
+    }
+    uint8_t * elemValidP = reinterpret_cast<uint8_t*>(valP) +
+        (sizeof(O)*8*len + 7) / 8 + index / 8;
+    bool elemValid = (*elemValidP & (0x1 << (index%8))) != 0x0;
+    if (elemValid) {
+        return thrust::make_tuple<O, bool>(*(valP + index), true);
+    } else {
```
jshencode: nit: no need for else

query/functor.hpp
voyager-dw: yes, the negative index will be reverse order, -1 is the last

PR #296, opened 2019-09-04
fix test case failure

voyager-dw:


datanode/bootstrap/bootstrap_server_test.go
```diff
@@ -58,6 +59,11 @@ var _ = ginkgo.Describe("bootstrap server", func() {
 	}
 
 	ginkgo.BeforeEach(func() {
+		testCreateTime := int64(1564646400)
```
shz117: you can use  SetCurrentTime instead:
```
utils.SetCurrentTime(time.Unix(1564646400, 0))
```

voyager-dw: that is not expected,  we still need the time difference between calls

shz117: I'm not following.. 
utils.Now() will always return time as represented by timestamp `1564646400` after your change.
which is the same as `utils.SetCurrentTime`'s behavior.

>> here are other calls to use the time functions
that's ok since we `utils.ResetClockImplementation()` after each test below.

voyager-dw: i mean we dont want to always return the fixed timestamp, we want to return the timestamp, plus the time elapsed between calls

PR #298, opened 2019-09-05
time duration instead of second since client side is sending duration directly

jshencode:
this is causing server side in non purge mode essentially forever when client side crashed


jshencode:
coverage does not make any sense. does not add or remove and line

PR #299, opened 2019-09-05
[NLJ] implementation layout

shz117:
- add comments for existing many to one join logic
 - add TODOs for code change to support NLJ
 - add query context fields needed for NLJ
 - compiler changes for matching many to many joins


voyager-dw:
Let's have some discussion first

PR #300, opened 2019-09-10
Array support null value

voyager-dw:
support null value for array

### DIFF
--- a/memstore/common/data_value.go
+++ b/memstore/common/data_value.go
@@ -23,6 +23,7 @@ import (
 	"strconv"
 	"strings"
 	"unsafe"
+	"math"
 )
 
 // NullDataValue is a global data value that stands a null value where the newly added
@@ -31,6 +32,12 @@ var NullDataValue = DataValue{}
 
 // SizeOfGeoPoint is the size of GeoPointGo in memory
 const SizeOfGeoPoint = unsafe.Sizeof(GeoPointGo{})
+// ZeroLengthArrayFlag is the value to represent 0 length array value
+// used in offset part of OffsetLength vector
+// when length part is non-zero value, the offset is real offset in memory
+// when length is 0 and offset is 0, the array value is invalid (by default)
+// when length is 0 and offset is ZeroLengthArrayFlag, this is a 0 length array value
+const ZeroLengthArrayFlag = math.MaxUint32
 
 // CompareFunc represents compare function
 type CompareFunc func(a, b unsafe.Pointer) int
@@ -300,7 +307,7 @@ func (v1 DataValue) ConvertToHumanReadable(dataType DataType) interface{} {
 			arrVal := make([]interface{}, num)
 
 			for i := 0; i < int(num); i++ {
-				if reader.IsValid(i) {
+				if reader.IsItemValid(i) {
 					var dataValue DataValue
 					if reader.itemType == Bool {
 						dataValue = DataValue{
@@ -715,11 +722,18 @@ type ArrayValueReader struct {
 
 // NewArrayValueReader is to create ArrayValueReader to read from upsertbatch, which includes the item number
 func NewArrayValueReader(dataType DataType, value unsafe.Pointer) *ArrayValueReader {
-	length := *((*uint32)(value))
-	return &ArrayValueReader{
-		itemType: GetElementDataType(dataType),
-		value:    unsafe.Pointer(uintptr(value) + 4),
-		length:   int(length),
+	if value == nil || uintptr(value) == 0 {
+		return &ArrayValueReader{
+			itemType: GetElementDataType(dataType),
+			value:    nil,
+			length:   int(0),
+		}
+	} else {
+		return &ArrayValueReader{
+			itemType: GetElementDataType(dataType),
+			value:    unsafe.Pointer(uintptr(value) + 4),
+			length:   int(*((*uint32)(value))),
+		}
 	}
 }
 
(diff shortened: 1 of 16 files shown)


shz117:
can you document list type data format for both live and archive batches in the [wiki](https://github.com/uber/aresdb/wiki/Vector-Store)? 
It should help review and external OSS community to understand.

memstore/list/archive_vector_party.go
```diff
@@ -139,16 +140,19 @@ func (vp *ArchiveVectorParty) GetDataValueByRow(row int) common.DataValue {
...
-		vp.SetOffsetLength(row, unsafe.Pointer(&zero), unsafe.Pointer(&zero))
+		vp.SetOffsetLength(row, unsafe.Pointer(&zero), unsafe.Pointer(&zero), false)
 		if row >= vp.lengthFilled {
 			vp.lengthFilled++
 		}
 		return
 	}
-	newLen := *(*uint32)(val)
+	var newLen uint32
+	if uintptr(val) != 0 {
```
jshencode: what does it mean when valid is true and val ptr is 0

memstore/common/data_value.go
```diff
@@ -32,6 +33,13 @@ var NullDataValue = DataValue{}
 // SizeOfGeoPoint is the size of GeoPointGo in memory
 const SizeOfGeoPoint = unsafe.Sizeof(GeoPointGo{})
 
+// ZeroLengthArrayFlag is the value to represent 0 length array value
+// used in offset part of OffsetLength vector
+// when length part is non-zero value, the offset is real offset in memory
+// when length is 0 and offset is 0, the array value is invalid (by default)
+// when length is 0 and offset is ZeroLengthArrayFlag, this is a 0 length array value
```
jshencode: seems like we just need offset to be nonzero right for this case?

memstore/list/archive_vector_party.go
jshencode: nvm, got it, but actually val != nil should be enough

memstore/list/live_vector_party.go
```diff
@@ -287,17 +289,19 @@ func (vp *LiveVectorParty) SetValue(row int, val unsafe.Pointer, valid bool) {
 
 	var newLen int
 	if valid {
-		newLen = int(*(*uint32)(val))
+		if uintptr(val) != 0 {
```
jshencode: same, val != nil should be better

memstore/list/live_vector_party.go
```diff
@@ -337,9 +341,11 @@ func (vp *LiveVectorParty) GetValue(row int) (val unsafe.Pointer, validity bool)
 	vp.RLock()
 	defer vp.RUnlock()
 
-	offset, length := vp.GetOffsetLength(row)
-	if offset == 0 && length == 0 {
+	offset, length, valid := vp.GetOffsetLength(row)
+	if !valid {
 		return unsafe.Pointer(uintptr(0)), false
+	} else if length == 0 {
+		return unsafe.Pointer(uintptr(0)), true
```
jshencode: i think we can just return nil, true

memstore/list/vector_party.go
```diff
@@ -38,19 +38,31 @@ type baseVectorParty struct {
...
 	length = *(*uint32)(vp.offsets.GetValue(2*row + 1))
+	valid = !(length == 0 && offset == 0)
 	return
 }
 
 // SetOffsetLength update offset/length for nth fow
-func (vp *baseVectorParty) SetOffsetLength(row int, offset, length unsafe.Pointer) {
-	vp.offsets.SetValue(2*row, offset)
-	vp.offsets.SetValue(2*row+1, length)
+func (vp *baseVectorParty) SetOffsetLength(row int, offset, length unsafe.Pointer, valid bool) {
```
jshencode: seems we can get validity from offset and length already, no need for the extra bool argument

query/iterator.hpp
```diff
@@ -421,13 +421,13 @@ class ArrayVectorPartyIterator
...
     if (length == 0) {
-        return thrust::make_tuple(reinterpret_cast<Value*>(NULL), true);
-    } else if (length == UINT32_MAX) {
+      if (offset == 0) {
         return thrust::make_tuple(reinterpret_cast<Value*>(NULL), false);
-    } else {
-        return thrust::make_tuple(
-                reinterpret_cast<Value*>(valuePtr + offset), true);
+      }
+      return thrust::make_tuple(reinterpret_cast<Value*>(NULL), true);
```
jshencode: can be simplified to return thrust::make_tuple(reinterpret_cast<Value*>(NULL), offset == 0);

memstore/common/data_value.go
voyager-dw: offset is not enouth, there is real valid cases when offset is 0

jshencode: no i mean when length is 0 we just need to know whether offset is zero or not to determine invalid or not, right?

voyager-dw: that is correct

memstore/common/data_value.go
```diff
@@ -715,11 +723,17 @@ type ArrayValueReader struct {
 
 // NewArrayValueReader is to create ArrayValueReader to read from upsertbatch, which includes the item number
 func NewArrayValueReader(dataType DataType, value unsafe.Pointer) *ArrayValueReader {
-	length := *((*uint32)(value))
+	if value == nil || uintptr(value) == 0 {
```
jshencode: is value == nil enough?

voyager-dw: good catch, cleaned up

PR #301, opened 2019-09-10
fix build broken due to merge

voyager-dw:


shz117:
same as #297 ?

voyager-dw:
yeah, not sure it happened again

PR #302, opened 2019-09-12
support manual bootstrap

lxning:
1. add url:bootstap.retry to accept manually bootstrap
2. add bootstrap retry button
3. add log in schema fetch


datanode/datanode.go
```diff
@@ -268,6 +268,18 @@ func (d *dataNode) startDebugServer() {
 
 	d.opts.InstrumentOptions().Logger().Infof("Starting HTTP server on dbg-port %d", d.opts.ServerConfig().DebugPort)
 	d.opts.InstrumentOptions().Logger().Fatal(http.ListenAndServe(fmt.Sprintf(":%d", d.opts.ServerConfig().DebugPort), debugRouter))
+
+	for {
```
jshencode: actually this will never get executed

also this for loop will block on the select

datanode/datanode.go
```diff
@@ -717,3 +719,18 @@ func mixedHandler(grpcServer *grpc.Server, httpHandler http.Handler) http.Handle
...
+		select {
+		case <-d.handlers.debugHandler.GetBootstrapRetryChan():
+			go func() {
+				err := d.bootstrapManager.Bootstrap()
+				if err != nil {
+					d.opts.InstrumentOptions().Logger().With("error", err.Error()).Error("error while retry bootstrapping")
+				}
+			}()
+		default:
```
jshencode: adding default here will create busy looping

PR #307, opened 2019-09-21
Array support 6 UUID

voyager-dw:
add uuid/geopoint type support in dimension and scratchspace/binder


.travis.yml
```diff
@@ -37,8 +37,8 @@ install:
   - ./build/cmake/bin/cmake .
   - .travis/install_golang_tools.sh
 script:
-  - make lint -j
-  - make travis
+  - travis_wait 30 make lint -j
```
jshencode: curious what is wait for?

query/aql_compiler.go
```diff
@@ -766,14 +766,22 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
 				e.RHS = expr.Cast(e.RHS, highestType)
 			}
 
-			if rhs != nil && lhs.DataType == memCom.GeoPoint {
+			if rhs != nil && e.LHS.Type() == expr.GeoPoint {
```
jshencode: why can't we just use lhs.DataType

query/aql_compiler.go
```diff
@@ -766,14 +766,22 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
...
 				if val, err := memCom.GeoPointFromString(rhs.Val); err != nil {
 					qc.Error = err
 				} else {
 					e.RHS = &expr.GeopointLiteral{
 						Val: val,
 					}
 				}
+			} else if rhs != nil && e.LHS.Type() == expr.UUID {
```
jshencode: same here

query/aql_compiler.go
```diff
@@ -1108,6 +1116,20 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
...
+				case memCom.UUID:
+					strLiteral, ok := secondArg.(*expr.StringLiteral)
+					if !ok {
+						qc.Error = utils.StackError(nil, "array function %s needs uuid string literal", e.Name)
+						break
+					}
+					if val, err := memCom.UUIDFromString(strLiteral.Val); err != nil {
+						qc.Error = err
+						break
+					} else {
```
jshencode: nit: else seems unnecessary

query/iterator.hpp
```diff
@@ -565,7 +566,14 @@ inline typename ConstantIterator<UUIDT>::type make_constant_iterator(
   return thrust::make_constant_iterator(
       thrust::make_tuple<UUIDT, bool>(defaultValue, defaultNull));
 }
-
+/*
+template <>
+inline typename ConstantIterator<GeoPointT>::type make_constant_iterator(
```
jshencode: remove if not needed

.travis.yml
voyager-dw: updated with new method, this intends to mitigate timeout issue

query/aql_compiler.go
voyager-dw: lhs is nil sometimes

query/aql_compiler.go
voyager-dw: lhs is not always var type, so it could be nil

.travis.yml
jshencode: might need to check how much slow down in the c code compilation

voyager-dw: yeah, had a ticket created to improve it, most time is on the scratch_transform and dimension_transform, as we added output types from 3 to 5.

PR #308, opened 2019-09-24
broker bug fix: 1. send rewritten query to datanode 2. return empty hll result correctly 3. resultSize is not len(result)

jshencode:
1. send rewritten query to data node which will rewrite enum string value to number
2. return empty hll result correctly
3. resultSize is not len(result)
4. rewritten dim not passed back

### DIFF
--- a/broker/query_plan_agg.go
+++ b/broker/query_plan_agg.go
@@ -451,7 +451,7 @@ func buildSubPlan(agg common.AggType, qc QueryContext, assignments map[topology.
 	root := NewMergeNode(agg)
 	for host, shardIDs := range assignments {
 		// make deep copy
-		currQ := *qc.AQLQuery
+		currQ := qc.GetRewrittenQuery()
 		for _, shard := range shardIDs {
 			currQ.Shards = append(currQ.Shards, int(shard))
 		}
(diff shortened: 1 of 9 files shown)


broker/query_plan_agg.go
```diff
@@ -451,7 +457,7 @@ func buildSubPlan(agg common.AggType, qc QueryContext, assignments map[topology.
 	root := NewMergeNode(agg)
 	for host, shardIDs := range assignments {
 		// make deep copy
-		currQ := *qc.AQLQuery
+		currQ := qc.GetRewrittenQuery()
```
voyager-dw: can this line move to outside of the loop?

broker/query_plan_non_agg.go
```diff
@@ -116,8 +116,8 @@ func NewNonAggQueryPlan(qc *QueryContext, topo topology.HealthTrackingDynamicTop
 	plan.nodes = make([]*StreamingScanNode, len(assignment))
 	i := 0
 	for host, shards := range assignment {
-		// make deep copy
-		currQ := *qc.AQLQuery
+		// get rewritten query after compilation
+		currQ := qc.GetRewrittenQuery()
```
voyager-dw: same here

broker/query_plan_agg.go
jshencode: not sure. its address is passed into the currQc, so I don't want to mess with that.

voyager-dw: likely make a copy of currQ instead of recalculation

voyager-dw:
lgtm

PR #310, opened 2019-09-27
schema validation update for array type

voyager-dw:


metastore/common/errors.go
```diff
@@ -77,4 +77,6 @@ var (
 	ErrHLLColumnDoesNotAllowDefaultValue = errors.New("hll column does not allow default value")
 	ErrInvalidTableBatchSize             = errors.New("Table batch size should be larger than zero")
 	ErrInvalidPrimaryKeyBucketSize       = errors.New("Table primary key bucket size should be larger than zero")
+	ErrInvalidPrimaryKeyDataType         = errors.New("Array data type can not used as primary key")
```
shz117: error msg could be more generic, to match error type - we could have other types that are not eligible as PK in the future.

metastore/validator.go
```diff
@@ -142,6 +142,10 @@ func (v tableSchemaValidatorImpl) validateIndividualSchema(table *common.Table,
 		if colIdDedup[colId] {
 			return common.ErrDuplicatedColumn
 		}
+		colDataType :=  memCom.DataTypeFromString(table.Columns[colId].Type)
+		if colDataType >= memCom.ArrayBool && colDataType <= memCom.ArrayInt64 {
```
shz117: maybe reuse `IsArrayType()`?

voyager-dw: nice catch, forget this function myself :)

metastore/validator.go
```diff
@@ -161,6 +165,10 @@ func (v tableSchemaValidatorImpl) validateIndividualSchema(table *common.Table,
 			if colIdDedup[sortColumnId] {
 				return common.ErrDuplicatedColumn
 			}
+			colDataType :=  memCom.DataTypeFromString(table.Columns[sortColumnId].Type)
+			if colDataType >= memCom.ArrayBool && colDataType <= memCom.ArrayInt64 {
```
shz117: `memCom.IsArrayType(colDataType)`

PR #311, opened 2019-09-27
add manual bootstrap

lxning:


api/debug_handler.go
```diff
@@ -725,3 +728,18 @@ func (handler *DebugHandler) ReadBackfillQueueUpsertBatch(w http.ResponseWriter,
 	common.RespondWithJSONObject(w, response)
 	return
 }
+
+// Bootstrap will turn on bootstrap based on the request.
+func (handler *DebugHandler) BootstrapRetry(w http.ResponseWriter, r *http.Request) {
```
voyager-dw: need to prevent bootstrap again if bootstrap is running or submitted

lxning: it is protected in bootstrap manager.

PR #312, opened 2019-09-30
hll memory estimate in MB instead of Bytes

jshencode:
this drastically under estimate memory usage for hll query and cause out of memory situation during massive hll query hits, such as weekly hll contracts

### DIFF
--- a/query/aql_processor.go
+++ b/query/aql_processor.go
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	hllQueryRequiredMemoryInMB = 10 * 1024
+	hllQueryRequiredMemoryInBytes = 10 * (1 << 30)
 )
 
 // batchTransferExecutor defines the type of the functor to transfer a live batch or a archive batch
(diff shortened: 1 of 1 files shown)


query/aql_processor.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	hllQueryRequiredMemoryInMB = 10 * 1024
+	hllQueryRequiredMemoryInBytes = 10 * (1 << 30)
```
voyager-dw: so originally we only reserved 10K for HLL? and now 10GB, isn't a little too bigger?

jshencode: yeah we have been reserved 10K.
10GB is to make sure each card only run one hll at a time since we cannot estimate hll effectively

voyager-dw: sure, we better do 2 observation:
1. how many real gpu memory allocated when running weekly hll contractor in production/staging
2. how it impacts production query when hll runs with 10GB reserved. Now as we under-reserve the memory, so other queries may still able to run. After we bump to 10G, other queries in the same machine will be totally blocked even the GPU might have spare space.

shz117 (approved):

PR #313, opened 2019-10-01
Array enum client

voyager-dw:
- enum array client ingestion support
- enum array debug batch display


metastore/schema_fetch.go
```diff
@@ -201,7 +201,7 @@ func (j *SchemaFetchJob) FetchEnum() {
 			continue
 		}
 		for _, column := range table.Columns {
-			if !column.IsEnumColumn() {
+			if !column.IsEnumBasedColumn() {
```
shz117: `!column.IsEnumBasedColumn() and !column.ISEnumArrayColumn()` ?

voyager-dw: ! IsEnumBasedColumn already covered both cases

PR #318, opened 2019-10-03
create seperate worker pools for query and ingestion

jshencode:


api/data_handler.go
```diff
@@ -64,11 +67,16 @@ func (handler *DataHandler) PostData(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	err = handler.memStore.HandleIngestion(postDataRequest.TableName, postDataRequest.Shard, upsertBatch)
-	if err != nil {
-		common.RespondWithError(w, err)
-		return
-	}
+	available := handler.workerPool.GoIfAvailable(func() {
```
voyager-dw: for ingestion, better do throttling instead of error out, or need to change the client to wait/retry

api/query_handler.go
```diff
@@ -85,7 +88,14 @@ func (handler *QueryHandler) HandleAQL(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	handler.handleAQLInternal(aqlRequest, w, r)
+	available := handler.workerPool.GoIfAvailable(func() {
```
voyager-dw: for query, better use some timeout also

api/data_handler.go
jshencode: client already does retry

api/query_handler.go
jshencode: no need to do timeout in such case, this is the same behavior as what we are doing currently on limit serve, only difference is separate query and ingestion

PR #321, opened 2019-10-10
force time column uint32 

lxning:


subscriber/common/message/message_parser.go
```diff
@@ -151,6 +151,16 @@ func (mp *Parser) CheckTimeColumnExistence(schema *metaCom.Table, columnDict map
 
 		columnID := columnDict[columnName]
 		if columnID == 0 && row[id] != nil {
+			// force time column value is uint32
```
voyager-dw: maybe better to check/update in client side code

lxning:
> since we are doing this change.
> I think it might be better to provide more supported underlying transformation for time column.
> Like from string (ISO timestamp format) to uint32 seconds will be quite useful to users

Transformation needs define the mapping b/w from item in message and to column in table schema. This definition has not been supported in RTA yet.

lxning:
merge with backfill support

PR #324, opened 2019-10-14
force time column value is uint32

lxning:


memstore/common/upsert_batch_builder.go
```diff
@@ -62,7 +63,15 @@ func (c *columnBuilder) SetValue(row int, value interface{}) error {
 	} else {
 		var err error
 		c.values[row], err = ConvertValueForType(c.dataType, value)
-		if err != nil {
+		if c.columnID == 0 && err != nil {
```
voyager-dw: how about check only when following combinations: 
1. fact table
2. first column with type uint32
3. overflow error

lxning: there is an overflow if ConvertValueForType returns error.

PR #326, opened 2019-10-15
Enable JSON compression

shz117:
Enabling gzip on json response is beneficial.
 - Great reduction in network usage (around 3x at 8MB), this should pay off even more when the 
 cluster is under pressure, or payload is larger
 - Throughput increase grows as payload size increases
  

  -  ~30% boost at payload size ~8MB

 - When payload is small, compressing actually increases size, but throughput not impacted much
 - Under “BestSpeed” gzip option, CPU overhead seems neglectable.
 - On per query level, avg response time reduced as well


api/common/response.go
```diff
@@ -73,27 +79,42 @@ func RespondBytesWithCode(w http.ResponseWriter, code int, bs []byte) {
...
+	if willCompress {
+		w.Header().Set(utils.HTTPContentEncodingHeaderKey, utils.HTTPContentEncodingGzip)
+	}
 	w.WriteHeader(code)
 	if jsonBytes != nil {
-		w.Write(jsonBytes)
+		if willCompress {
+			gw, err = gzip.NewWriterLevel(w, gzip.BestSpeed)
+			if err != nil {
+				RespondWithError(w, ErrFailedToJSONMarshalResponseBody)
```
jshencode: return

PR #327, opened 2019-10-18
fix array device test

voyager-dw:


query/functor_unittest.cu
```diff
@@ -1175,33 +1175,38 @@ TEST(ArrayLengthTest, CheckArrayLengthFunctor) {
                          3, 1, 2, 3, 0x07, 0,
                          1, 1, 0x01, 0};
 
-  int expectedVals[6] = {2, 1, 3, 0, 0, 1};
-
+  uint32_t expectedVals[6] = {2, 1, 3, 0, 0, 1};
+  bool expectedNulls[6] = {true, true, true, false, true, true};
+#ifdef RUN_ON_DEVICE
```
shz117: I wonder why we didn't need this in the tests before. does this mean the other functor tests only run on host?

voyager-dw: correct. for those tests called through UnaryTransform/BinaryTransform, they are testing though device, otherwise, they are just testing on host.

shz117: i see 👍

PR #329, opened 2019-10-22
block int64 for binary transformations

shz117:
int64 type can not be used for binary transformations for now.
currently query fail until c++ type binding phase, add validation logic to fail those in compilation time.

### DIFF
--- a/query/aql_compiler.go
+++ b/query/aql_compiler.go
@@ -1496,6 +1496,14 @@ func (qc *AQLQueryContext) processFilters() {
 				geoFilterFound = true
 			}
 		} else {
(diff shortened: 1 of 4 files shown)


query/aql_compiler.go
```diff
@@ -678,6 +687,11 @@ func (qc *AQLQueryContext) Rewrite(expression expr.Expr) expr.Expr {
 			return expression
 		}
 
+		if err := blockInt64(e.LHS, e.RHS); err != nil {
```
jshencode: maybe a TODO here to mark that we currently block binary operation on int64 types

PR #336, opened 2019-10-31
purge metastore and diskstore table shard data

jshencode:
before adding new table shard and after removing table shard

### DIFF
--- a/memstore/memstore.go
+++ b/memstore/memstore.go
@@ -45,7 +45,7 @@ type MemStore interface {
 	// GetHostMemoryManager returns the host memory manager
 	GetHostMemoryManager() common.HostMemoryManager
 	// AddTableShard add a table shard to the memstore
-	AddTableShard(table string, shardID int, needPeerCopy bool)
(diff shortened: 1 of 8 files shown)


memstore/memstore.go
```diff
@@ -281,6 +281,17 @@ func (m *memStoreImpl) AddTableShard(table string, shardID int, needPeerCopy boo
 		if needPeerCopy {
 			tableShard.needPeerCopy = 1
 		}
+
+		// purge to make sure disk space is clean for new table shard when it is added
+		if needPurge {
+			if err := tableShard.diskStore.DeleteTableShard(table, shardID); err != nil {
+				utils.GetLogger().With("table", table, "shard", shardID, "error", err.Error()).Fatalf("failed to purge table shard data")
```
shz117: is logging enough?
maybe emit metrics so we can create alerts on it.

jshencode: this will be fatal and crash the server

shz117: oops, didn't realize it's Fatal..

PR #338, opened 2019-11-06
fix array table creation and null value length issue

voyager-dw:


metastore/disk_metastore.go
```diff
@@ -719,7 +719,7 @@ func (dm *diskMetaStore) ExtendEnumDict(table, columnName string, enumCases []st
 	newEnumID := len(existingCases)
 
 	enumCardinalityLimit := 1 << 8
-	if column.Type == common.BigEnum {
+	if column.Type == common.BigEnum || column.Type == common.ArrayBigEnum {
```
jshencode: can use EnumCardinality?

PR #341, opened 2019-11-12
add array integration tests

voyager-dw:
add integration test package and add tests for array


cmd/aresd/cmd/cmd.go
```diff
@@ -56,6 +57,47 @@ type Options struct {
...
+	options *Options
+	// server is http server instance started inside AresD
+	server *http.Server
+	// WaitCond is used to notify other route that the server is started
+	WaitCond *sync.Cond
+}
+
+// NewAresD create singleton of AresD
+func NewAresD(cfg common.AresServerConfig, options *Options) *AresD {
+	once.Do(func() {
```
jshencode: if there won't be multiple threads accessing this method, we don't need to make this a singleton

if only in test suites we will have multi threads sharing one aresd then the singletone should be placed in test suite only

utils/http.go
```diff
@@ -164,3 +164,24 @@ func LimitServe(port int, handler http.Handler, httpCfg common.HTTPConfig) {
 	}
 	GetLogger().Fatal(server.Serve(listener))
 }
+
+// LimitServe will start a http server on the port with the handler and at most maxConnection concurrent connections.
+func LimitServeImmediateReturn(port int, handler http.Handler, httpCfg common.HTTPConfig) (chan error, *http.Server) {
```
jshencode: nit: maybe LimitServeAsync?

cmd/aresd/cmd/cmd.go
voyager-dw: this is just the prevention for future use to create multiple instance

jshencode: I think whoever is the user of NewAresD method should control how many instance they want, not the NewAresD method itself.

In this case, the test suite is the user of NewAresD

voyager-dw: suppose only one aresd can run, while it will need more logic to control it. I will rollback to make it simpler

PR #343, opened 2019-11-18
stop logging result body

shz117:


shz117:
not needed anymore with #344

PR #347, opened 2019-12-04
Consolidate http utils

jshencode:


utils/http.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	HTTPContentTypeHeaderKey     = "content-type"
+	HTTPContentTypeHeaderKey     = "Content-Type"
 	HTTPAcceptTypeHeaderKey      = "accept"
```
shz117: nit: might as well capitalize this

PR #348, opened 2019-12-06
initial change

voyager-dw:


query/context/query_context_helper.go
```diff
@@ -0,0 +1,743 @@
...
+	"github.com/uber/aresdb/query/common"
+	"github.com/uber/aresdb/query/expr"
+	"github.com/uber/aresdb/utils"
+	"strconv"
+	"strings"
+)
+
+// QueryContextHelper is a helper Class to group common code
+// for broker query compiler and datanode query compiler
+type QueryContextHelper struct {
```
shz117: I feel `QueryCompilerContext` is a better name: it contains all data and methods for compiling AQL queries.

voyager-dw: just class name or more restructure?

